PR #66, opened 2024-01-02
Provide an ABI to enable projects written in other languages to use nigiri as a lib

traines-source:
As discussed previously, this PR enables a limited subset of nigiri to be used by other non C++ code. This is just a draft as a basis for further discussion. I have no experience in C++ whatsoever, so please review carefully.

As an example, the wrapping Rust crate(s) I have created are available at [traines-source/motis-nigiri-rust](https://github.com/traines-source/motis-nigiri-rust).

A couple of decisions/questions that need discussion:
1. I have decided to create a pure C wrapper that then can be wrapped by the C-ABI compatible language of your choice. While it is possible to have other programming languages interface with C++ directly, this is usually not recommended (see e.g. [here](https://stackoverflow.com/a/75556744/1666456) and [here](https://docs.rust-embedded.org/book/interoperability/rust-with-c.html#building-a-c-api)). The C++ ABI might not be as stable and in particular you pull all sorts of things from the C++ stdlib into your interface. To a certain extent, these types can be treated as opaque, but it's a huge hassle. Rust-specific alternatives like [cxx](https://github.com/dtolnay/cxx) can do this, but I think this would make it more difficult (or impossible?) to use the ABI from other languages.
2. Instead of providing getters for everything, the functions return (pointers to) structs.
3. The caller is responsible to call the respective `*_destroy_*` functions after being done. The nigiri lib then properly frees the memory (see a similar approach [here](https://github.com/google-deepmind/open_spiel/commit/669c3234376c79b6125fee6ddaa7ca7aa0efe409)). I chose this approach originally because some struct fields might actually be pointers to nigiri-internal data to avoid copying. But I think by now, most things are copied anyways, for a variety of reasons (e.g. strings need to be copied to be null-terminated, alternative: separate length field).
4. This unfortunately means that most things are copied at least twice (in the C wrapper and most likely in the calling code).
5. The nigiri lib has loads of dependencies. It is quite a hassle to link all of them in the right order without the help of CMake. In my Rust wrapper, I grep through the verbose output of `ninja` to find them, [this is](https://github.com/traines-source/motis-nigiri-rust/blob/master/nigiri-sys/build.sh#L6) [quite hacky](https://github.com/traines-source/motis-nigiri-rust/blob/master/nigiri-sys/build.rs#L12). I wonder whether there is a better solution (but [I'm not the first one](https://stackoverflow.com/questions/74851275/rust-cargo-cmake-integration-with-c-library-dependencies)). Maybe one could try to build an alternative version of the nigiri lib that statically links all dependencies, but this doesn't seem to be very straightforward either.
6. I decided to expose transports that are not yet exploded by days. This enables callers to leverage the bitfield advantages. However, the bitfield is not transferred in the `nigiri_transport` struct (this would be quite cumbersome in pure C), instead, the function `nigiri_is_transport_active` is provided.
7. How do we handle transports added in RT? Do we extend the `nigiri_event_change` struct to cover all necessary information? Or do we just add an `rt_added` flag to that struct and a function `nigiri_get_rt_transport` that allows to retrieve from the `rt_timetable` but otherwise has the same signature as `nigiri_get_transport`?
8. Why does a route have multiple claszes (`timetable.route_section_clasz_`)? Is this to allow for trains that are e.g. both RE and IC? Currently just taking the first.
9. Commit `f13753b` contains some unrelated fixes of things that I stumbled upon. 

TODOs:
- [x] in/out_allowed
- [ ] RT: handle track changes, cancellations/skipped, transports only in RT, possibly more RT special cases
- [ ] expose footpaths
- [ ] extend structs with further relevant fields (?)
- [ ] verify correctness of timestamp handling (currently, caller is responsible for resolution of base_day + day_idx + event_mam)
- [ ] error handling
- [x] tests (there are some tests in [motis-nigiri-rust](https://github.com/traines-source/motis-nigiri-rust))
- [ ] passive initialization. Currently, the caller has to initiate loading of the timetable. For other usecases, it might be useful to register some form of callback to be able to use a timetable that someone else (the main motis process) is loading.

### DIFF
--- a/include/nigiri/abi.h
+++ b/include/nigiri/abi.h
@@ -11,7 +11,7 @@ typedef struct nigiri_timetable nigiri_timetable_t;
 struct nigiri_transport {
   // bitfield_idx_t bitfield_idx_;
   uint32_t route_idx;
-  uint32_t n_event_mams;
+  uint16_t n_event_mams;
   int16_t* event_mams;
   const char* name;
 };
@@ -22,15 +22,15 @@ struct nigiri_stop {
   char* name;
   double lon;
   double lat;
-  uint32_t transfer_time;
-  int32_t parent;
+  uint16_t transfer_time;
+  uint32_t parent;
 };
 typedef struct nigiri_stop nigiri_stop_t;
 
 struct nigiri_route {
-  uint32_t n_stops;
+  uint16_t n_stops;
   uint32_t* stops;
-  uint32_t clasz;
+  uint16_t clasz;
 };
 typedef struct nigiri_route nigiri_route_t;
 
--- a/include/nigiri/abi.h
+++ b/include/nigiri/abi.h
@@ -30,9 +30,9 @@ struct nigiri_location {
 typedef struct nigiri_location nigiri_location_t;
 
 struct nigiri_route_stop {
-  uint32_t location_idx : 30;
-  bool in_allowed : 1;
-  bool out_allowed : 1;
+  unsigned int location_idx : 30;
+  unsigned int in_allowed : 1;
+  unsigned int out_allowed : 1;
 };
 typedef struct nigiri_route_stop nigiri_route_stop_t;
 
@@ -75,7 +75,9 @@ void nigiri_destroy_location(const nigiri_location_t* location);
 
 void nigiri_update_with_rt(const nigiri_timetable_t* t,
                            const char* gtfsrt_pb_path,
-                           void (*callback)(nigiri_event_change_t));
+                           void (*callback)(nigiri_event_change_t,
+                                            void* context),
+                           void* context);
 
 #ifdef __cplusplus
 }
--- a/include/nigiri/abi.h
+++ b/include/nigiri/abi.h
@@ -1,4 +1,5 @@
 #include <stdint.h>
+#include <stdbool.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -33,11 +34,22 @@ struct nigiri_route {
 };
 typedef struct nigiri_route nigiri_route_t;
 
+struct nigiri_event_change {
+    uint32_t transport_idx;
+    uint16_t day;
+    uint32_t stop_idx;
+    bool is_departure;
+    int16_t delay;
+    bool cancelled;
+};
+typedef struct nigiri_event_change nigiri_event_change_t;
+
 
 nigiri_timetable_t *nigiri_load(const char* path, int64_t from_ts, int64_t to_ts);
 void nigiri_destroy(const nigiri_timetable_t *t);
 uint32_t nigiri_get_transport_count(const nigiri_timetable_t *t);
 nigiri_transport_t *nigiri_get_transport(const nigiri_timetable_t *t, uint32_t idx);
+bool nigiri_is_transport_active(const nigiri_transport_t *transport, uint16_t day);
 void nigiri_destroy_transport(const nigiri_transport_t *transport);
 nigiri_route_t *nigiri_get_route(const nigiri_timetable_t *t, uint32_t idx);
 void nigiri_destroy_route(const nigiri_route_t *route);
@@ -45,7 +57,7 @@ uint32_t nigiri_get_stop_count(const nigiri_timetable_t *t);
 nigiri_stop_t *nigiri_get_stop(const nigiri_timetable_t *t, uint32_t idx);
 void nigiri_destroy_stop(const nigiri_stop_t *stop);
 
-void update_with_rt(nigiri_timetable_t *t);
+void nigiri_update_with_rt(const nigiri_timetable_t *t, const char* gtfsrt_pb_path, void (*callback)(nigiri_event_change_t));
 
 #ifdef __cplusplus
 }
--- a/include/nigiri/abi.h
+++ b/include/nigiri/abi.h
@@ -17,7 +17,7 @@ struct nigiri_transport {
 };
 typedef struct nigiri_transport nigiri_transport_t;
 
-struct nigiri_stop {
+struct nigiri_location {
   char* id;
   char* name;
   double lon;
@@ -25,11 +25,18 @@ struct nigiri_stop {
   uint16_t transfer_time;
   uint32_t parent;
 };
-typedef struct nigiri_stop nigiri_stop_t;
+typedef struct nigiri_location nigiri_location_t;
+
+struct nigiri_route_stop {
+    uint32_t location_idx: 30;
+    bool in_allowed: 1;
+    bool out_allowed: 1;
+};
+typedef struct nigiri_route_stop nigiri_route_stop_t;
 
 struct nigiri_route {
   uint16_t n_stops;
-  uint32_t* stops;
+  nigiri_route_stop_t* stops;
   uint16_t clasz;
 };
 typedef struct nigiri_route nigiri_route_t;
@@ -59,9 +66,9 @@ bool nigiri_is_transport_active(const nigiri_timetable_t* t,
                                 uint16_t day_idx);
 nigiri_route_t* nigiri_get_route(const nigiri_timetable_t* t, uint32_t idx);
 void nigiri_destroy_route(const nigiri_route_t* route);
-uint32_t nigiri_get_stop_count(const nigiri_timetable_t* t);
-nigiri_stop_t* nigiri_get_stop(const nigiri_timetable_t* t, uint32_t idx);
-void nigiri_destroy_stop(const nigiri_stop_t* stop);
+uint32_t nigiri_get_location_count(const nigiri_timetable_t* t);
+nigiri_location_t* nigiri_get_location(const nigiri_timetable_t* t, uint32_t idx);
+void nigiri_destroy_location(const nigiri_location_t* location);
 
 void nigiri_update_with_rt(const nigiri_timetable_t* t,
                            const char* gtfsrt_pb_path,
--- a/include/nigiri/abi.h
+++ b/include/nigiri/abi.h
@@ -1,0 +1,51 @@
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct nigiri_timetable;
+typedef struct nigiri_timetable nigiri_timetable_t;
+
+struct nigiri_transport {
+    //bitfield_idx_t bitfield_idx_;
+    uint32_t route_idx;
+    uint32_t n_event_mams;
+    int16_t *event_mams;
+    const char* name;
+};
+typedef struct nigiri_transport nigiri_transport_t;
+
+struct nigiri_stop {
+    char *id;
+    char *name;
+    double lon;
+    double lat;
+    uint32_t transfer_time;
+    int32_t parent;
+};
+typedef struct nigiri_stop nigiri_stop_t;
+
+struct nigiri_route {
+    uint32_t n_stops;
+    uint32_t *stops;
+    uint32_t clasz;
+};
+typedef struct nigiri_route nigiri_route_t;
+
+
+nigiri_timetable_t *nigiri_load(const char* path, int64_t from_ts, int64_t to_ts);
+void nigiri_destroy(const nigiri_timetable_t *t);
+uint32_t nigiri_get_transport_count(const nigiri_timetable_t *t);
+nigiri_transport_t *nigiri_get_transport(const nigiri_timetable_t *t, uint32_t idx);
+void nigiri_destroy_transport(const nigiri_transport_t *transport);
+nigiri_route_t *nigiri_get_route(const nigiri_timetable_t *t, uint32_t idx);
+void nigiri_destroy_route(const nigiri_route_t *route);
+nigiri_stop_t *nigiri_get_stop(const nigiri_timetable_t *t, uint32_t idx);
+void nigiri_destroy_stop(const nigiri_stop_t *stop);
+
+void update_with_rt(nigiri_timetable_t *t);
+
+#ifdef __cplusplus
+}
+#endif
--- a/include/nigiri/abi.h
+++ b/include/nigiri/abi.h
@@ -1,5 +1,5 @@
-#include <stdint.h>
 #include <stdbool.h>
+#include <stdint.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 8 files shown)


felixguendling:
:wave: This looks already really nice!

Regarding the issues:

1. For other languages like NodeJS or Python, there often exist specific ways in C++ (like [pybind11](https://github.com/pybind/pybind11)) to create bindings. So I would not make the decision for Rust depend on how other programming languages *might* benefit the most. I would guess that the `cxx` Rust package would be updated accordingly in case the C++ ABI changes. In the past, the C++ ABI has been quite stable, so I would not consider ABI compatibility of the C++ standard library an issue for a Rust wrapper. However, I lack practical experience with both approaches (C wrapper / `cxx` crate), so whatever works is fine. I guess it's more important to have something that's convenient to use in Rust, which can be achieved with both approaches.
2. :+1: 
3. In terms of performance, copying should always be avoided. I'm not a Rust programmer, but the concept of a Slice looks a bit like a "not null terminated" string (aka `string_view` or `span` in C++). So I would guess that Rust can deal with data that lacks null-termination.
4. see (3)
5. In nigiri, all dependencies are already statically linked. So your last proposal should work. Let me know if you encounter any problems. As long as your approach with reading `ninja` verbose log works, that's also fine.
6. :+1: 
7. Currently, GTFS-RT does not support adding new trips. There are proposals to change this and also support added stops. However, at the moment there are no real additional services (that were not already contained in the schedule data) supported in GTFS-RT (and therefore also in nigiri).
8. Currently, this does not happen. It was a preparation for adding support for trains that are merged and split. Then you have several trains coupled together and therefore more than one clasz. Until this is implemented (which will probably require more changes to nigiri's timetable data model), it's totally fine to only select the first clasz.
9. Thank you! :smiley: 

Regarding TODOs: it might make sense to merge an initial version and then proceed with more PRs later.

Would it make sense to have some kind of testing / CI directly in the nigiri repository so nobody accidentally destroys the compatibility to the Rust code?

traines-source:
Sounds good, thanks for the quick response!

1. `cxx` will generate code "on both sides". Maybe pybind11 does a similar thing. But this means that you would have to add new code to nigiri for every new language. Using the pure C way, the defined ABI in nigiri is "authoritative" and anyone wanting to use it can generate bindings out of that for their language of choice. But if there were, say a std::string in that interface, this generation logic needed to somehow handle the conversion from a std::string to the string type of that language directly – or it just generates loads upon loads of bindings for all the methods etc. of std::string and the user has to worry about it manually (which is the case when using Rust's `bindgen`). So since you're fine with the pure C way, let's stick with that for the time being :)
2. -
3. There are actually multiple reasons why mostly I have to copy:
* event_mams: In nigiri, `route_stop_times_` are sorted by stop, but I want to return one transport at a time -> copy to new array
* stops: `route_location_seq_` is an array of stops that also contain in_allowed/out_allowed. Currently, I just transfer a list of location_idx -> copy needed. I would transfer in_allowed/out_allowed in a separate array (not yet implemented). When transferring everything in one array, one could probably avoid the copy by casting the stop::value_type as a pure C struct with matching fields and hope that everything lines up... Sounds ugly to me, but possible.
* And finally the strings. Rust certainly uses non-terminated strings, because (just as C++ I suppose) it internally stores the length. But how do we do it in the pure C ABI? We could alternatively add a separate length field next to each of the char ptrs. But since the natural C way is null-termination, it is actually easier to get a Rust string from a null-terminated char ptr than from a char ptr and some additional length :) Independently of that, copying on the Rust side is necessary, if you don't want to defer the `*_destroy_*`calls until forever or giving the caller (Rust) the complete responsibility to clean up which IMO is a bad idea, but idk. Of course, if we really achieved not copying at all (which I doubt), then we could just return ("copying" :D) structs instead of ptrs and need no `*_destroy_*` at all.
4. -
5. Mmh. As far as I understood, the libnigiri.a artifact does not contain all the dependencies, absl, utl, geo etc. I think this is the way it is meant to be (see e.g. [here](https://stackoverflow.com/a/14200728) for a workaround), since otherwise, when linking that library somewhere else (say in motis), you would possibly quadruple all these dependencies. An executable like nigiri-server on the other hand does contain all these static dependencies. At least that's my current level of confusion, and anyways the Rust build complains about undefined references as long as I don't indicate these dependencies. Or do I have a misconception somewhere?
6. -
7. I thought this is what the schedule relationship `ADDED` is for. I thought to have seen this somewhere in the nigiri code as well. But, on a side note, I am getting `unsupported schedule relationship ADDED` log entries when loading some [example swiss gtfsrt file](https://github.com/traines-source/motis-nigiri-rust/blob/master/nigiri-sys/tests/fixtures/2024-01-02T01_48_02%2B01_00.gtfsrt). Haven't investigated yet. 
8. Nice, less work :)
9. -
10. merging before TODOs: yes why not if you prefer that. But I think some things (like tests) should be done before merging, I'll hopefully get around to this tomorrow (pending possible changes to memory mgmt/copying, since this may change the interface quite a bit I think).
11. I don't think we should pull in any Rust-specific stuff. But some tests to just verify that the ABI is working would probably be good. Then at least one also has to change the tests :D Probably these would be quite similar to the one I've written in Rust in `nigiri-sys`.

felixguendling:
Unfortunately, I don't have any experience wrapping C or C++ code with Rust yet. So take everything I write with a grain of salt.

Since the Rust code for read access to `timetable` data doesn't modify the data and references are valid until the `timetable` struct gets deallocated (which only happens at program termination in most cases), an alternative to copying data would be to only point to the data in the timetable (which would require to have a Rust type mirroring `stop`). This would eliminate the need for copies/`*_destroy_*` and therefore enable a "zero cost abstraction". Therefore, regarding performance (execution time + memory usage) it would not matter if the timetable is used in Rust or in C++ (which would be amazing). For `rt_timetable` it's a different story since this `struct` currently gets replaced with each real-time update. So here, references in Rust should only be used until the next real-time update.

Regarding linking: would it be an option to just link all `*.a` files from the build folder?

The `ADDED` schedule relation ship is deprecated in favor of `DUPLICATE`. Both require the trip to be contained in the GTFS-static timetable. It will just be copied to the current day. So currently, there's no way to create a trip with a new stop sequence not already contained in the GTFS-static timetable.

Yes, tests would be nice to have for this PR already.

traines-source:
Well, I have even less experience with that :D

Maybe we're talking about different things, but I think the relevant thing is how to expose everything in the pure C interface. Rust will follow.
E.g. we would need to have a pure C struct to mirror the `stop` type. Am I correct that then you would use `reinterpret_cast` and hope that alignment and padding etc. line up? I suspect this will work but is this guaranteed?
And then, we would not offer an API for individual transports, but just give the caller a pointer to the backing arrays of vectors like `route_stop_times_` and `route_location_seq_` etc.
I think in most use cases (at least in my use case :) this doesn't change anything about the required copy on the caller's side, because the caller most likely does not want to use the same data structure as nigiri. What still disturbs me most is that there will be two copies of everything held in memory indefinitely. I think this is much less desirable than some temporary copies happening while iterating over the transports. So in the distant future, I still think it will be desirable to have an alternative working mode of this ABI, where nigiri will relay the transports to the caller while it is loading them  and never put them into its own data structures.

Linking: Linking everything from the build folder is roughly what I do. But the order is extremely relevant, that's why I have to grep the command `ninja` is running to build/link `nigiri-server`.

`ADDED`, ah, I didn't realize that the gtfs.org and Google documentation about GTFSRT differs that much. But wouldn't this already cover many usecases, like Deutsche Bahn cancelling trains, but adding a replacement train on the same route at the same time? Wonder if that's what it's used for in the Swiss GTFSRT feed as well.

traines-source:
So I have gotten rid of most of the copying now (of course apart from struct allocation and apart from event_mams, I don't see how this would be possible without copying, as long as we keep the transport-centric API). I'm happy :)
For that, I have introduced a mirroring stop type in the C interface. For the time being, it uses `reinterpret_cast` (see [this line](https://github.com/motis-project/nigiri/blob/2ca31c5523ac752cd8ce2a605f0529182d4e79d4/src/abi.cc#L160)), I don't see how this would be possible otherwise without copying (I don't really understand `bit_cast`, but doesn't it also copy?). At least it works, whether it is UB, idk.
I also added one big test.
Everything else should be fine then :)

felixguendling:
Valgrind and address sanitizer seem to be unhappy. Can you please check? :)

traines-source:
Ok, for some reason I thought that `delete` would not call the destructor...

And the use of `reinterpret_cast` seems indeed to be UB, at least I suspect that's why with msvc, the tests fail due to garbled stops. So I fear the only non-copy option would be to expose a pointer to the array of `uint32_t` as you said and have the caller worry about pulling apart the bits. I think that's undesirable, so I reverted to copying with `bit_cast` for the time being. Let's see.

felixguendling:
Looks like MSVC doesn't provide `std::bit_cast` yet. But it can easily be implemented with `std::memcpy`.

Linter errors:
```
/__w/nigiri/nigiri/test/abi_test.cc:333:12: error: variable 'test_event_change_counter' is non-const and globally accessible, consider making it const [cppcoreguidelines-avoid-non-const-global-variables,-warnings-as-errors]
static int test_event_change_counter = 0;
           ^
/__w/nigiri/nigiri/test/abi_test.cc:333:12: error: variable 'test_event_change_counter' declared 'static', move to anonymous namespace instead [misc-use-anonymous-namespace,-warnings-as-errors]
/__w/nigiri/nigiri/test/abi_test.cc:393:3: error: variable 'l0_name' of type 'std::string' (aka 'basic_string<char>') can be declared 'const' [misc-const-correctness,-warnings-as-errors]
  std::string l0_name(l0->name, l0->name_len);
  ^
              const 
/__w/nigiri/nigiri/test/abi_test.cc:398:3: error: variable 'l9_name' of type 'std::string' (aka 'basic_string<char>') can be declared 'const' [misc-const-correctness,-warnings-as-errors]
  std::string l9_name(l9->name, l9->name_len);
  ^
              const 
/__w/nigiri/nigiri/test/abi_test.cc:400:3: error: variable 'l9_id' of type 'std::string' (aka 'basic_string<char>') can be declared 'const' [misc-const-correctness,-warnings-as-errors]
  std::string l9_id(l9->id, l9->id_len);
  ^
              const 
/__w/nigiri/nigiri/test/abi_test.cc:409:3: error: variable 'l35_name' of type 'std::string' (aka 'basic_string<char>') can be declared 'const' [misc-const-correctness,-warnings-as-errors]
  std::string l35_name(l35->name, l35->name_len);
  ^
              const 
/__w/nigiri/nigiri/test/abi_test.cc:411:3: error: variable 'l35_id' of type 'std::string' (aka 'basic_string<char>') can be declared 'const' [misc-const-correctness,-warnings-as-errors]
  std::string l35_id(l35->id, l35->id_len);
  ^
              const 
/__w/nigiri/nigiri/test/abi_test.cc:420:3: error: variable 'l36_name' of type 'std::string' (aka 'basic_string<char>') can be declared 'const' [misc-const-correctness,-warnings-as-errors]
  std::string l36_name(l36->name, l36->name_len);
  ^
              const 
/__w/nigiri/nigiri/test/abi_test.cc:422:3: error: variable 'l36_id' of type 'std::string' (aka 'basic_string<char>') can be declared 'const' [misc-const-correctness,-warnings-as-errors]
  std::string l36_id(l36->id, l36->id_len);
  ^
              const 
/__w/nigiri/nigiri/test/abi_test.cc:443:3: error: variable 't_name' of type 'std::string' (aka 'basic_string<char>') can be declared 'const' [misc-const-correctness,-warnings-as-errors]
  std::string t_name(transport->name, transport->name_len);
  ^
              const 
```

When you change those lines, can you please convert them to [AAA style](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/)? Left side of assignment is always `auto const`, right side would be `std::string{ptr, len}`.

traines-source:
Fixed now hopefully. I haven't come up with a straightforward alternative for the global variable counting the callback invocations, so I added a NOLINT, since it's just a test. Lmk if you have a better idea.

felixguendling:
Since the counter and the test function is only used in one test case, it would be cleaner to have a lambda that captures the local counter variable by reference:

```cpp
TEST(/* ... */) {
   auto counter = 0U;
   auto const test_fn = [&]() { ++counter; /* ... */ };
   test_fn();
}
```

traines-source:
I had envisaged that, but apparently [it is not possible to pass a capturing lambda as a function pointer](https://stackoverflow.com/questions/28746744/passing-capturing-lambda-as-function-pointer) (frankly haven't verified it myself, but makes sense). But I could just expect a lambda in the intermediate interface which I have created anyways for the test to avoid having to pass a file path and loading the files from disk. In production code, the function pointer would then be wrapped by two lambdas or the test couldn't test the creation of the `nigiri_event_change_t` struct.

I think I also know the reason why with msvc, the test is still garbled. Not only `reinterpret_cast` is UB, but also the layout when using different types in a bitfield (I was naive and used two bools for in/out_allowed) is undefined, [on Windows this will apparently use more memory](https://stackoverflow.com/questions/308364/c-bitfield-packing-with-bools). So I thought:
* Great, let's use three `uint32_t` then, as in the original `stop` struct. Shoot. In C (not in C++), this is UB, as [only fundamental types (i.e. `unsigned int`) are allowed in a bitfield](https://www.ibm.com/docs/en/i/7.1?topic=declarations-bit-field-members)
* Great, let's use `unsigned int` then. Shoot. A bitfield [may not exceed the width of the type](https://stackoverflow.com/questions/14553632/maximum-size-of-a-bit-field-in-c-or-c), i.e. on 16-bit systems we again would have UB.
* Great, I give in and we just return `uint32_t`s and the caller can use some bitmask. Shoot. Actually the entire [bit-level layout of bitfields is implementation-dependent](https://en.cppreference.com/w/cpp/language/bit_field), so I think there is no sure way of knowing which bits to mask.

So, what do we do?
* Use three `uint32_t` for the bitfield, possibly even with `reinterpret_cast`, since both is UB, but most likely [actually works everywhere](https://community.sonarsource.com/t/change-type-of-bit-field-uint32-not-acceptable-type/26853)?
* Just copy to normal structs with normal, huge booleans or two/three separate arrays for location_idx, in_allowed, out_allowed?

This is all very interesting and I'm learning a lot, but it's starting to get tedious :D

felixguendling:
Usually, C APIs allow the user to pass a `void*` data pointer for those reasons.
Here's an example: https://github.com/tidwall/rtree.c/blob/master/rtree.h#L68

But if it's only for the test cases, a global variable is also acceptable. But it could make sense to add a user data pointer for better usability of the API. In case there's no user data, it can just be set to null and be ignored in the callback.

As `unsigned int` is the same as `std::uint32_t` on all supported platforms and compilers (MOTIS mainly targets x86_64 and aarch64 architecture, but even for x86 or arm this is true), it should be fine to use 3x `unsigned int` in the C API. Also implementation defined stuff is fine as long as it works on all targeted compilers (MSVC, Clang, GCC) for those platforms (x86_64, aarch64 - if x86 and arm cause problems, I would just drop them). Only UB would be problematic as the compiler behaviour can change drastically between versions when new optimizations are added.

traines-source:
That sounds good. Done. Let's see what will be red next :)

felixguendling:
Looks good! 👍

traines-source:
Cool. Thank you for your support! 🙌

PR #67, opened 2024-02-18
Add support for GTFS route colors

vkrause:
For consumption by MOTIS, useful for result presentation.

### DIFF
--- a/include/nigiri/loader/gtfs/route.h
+++ b/include/nigiri/loader/gtfs/route.h
@@ -21,6 +21,8 @@ struct route {
   std::string long_name_;
   std::string desc_;
   clasz clasz_;
+  color_t color_;
+  color_t text_color_;
 };
 
 using route_map_t = hash_map<std::string, std::unique_ptr<route>>;
--- a/include/nigiri/rt/frun.h
+++ b/include/nigiri/rt/frun.h
@@ -39,6 +39,8 @@ struct frun : public run {
     clasz get_clasz(event_type = event_type::kDep) const noexcept;
     clasz get_scheduled_clasz(event_type = event_type::kDep) const noexcept;
 
+    route_color get_route_color(event_type = event_type::kDep) const noexcept;
+
     bool in_allowed() const noexcept;
     bool out_allowed() const noexcept;
     bool is_canceled() const noexcept;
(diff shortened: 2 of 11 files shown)


felixguendling:
Would it make sense to have a test case to make sure this functionality doesn't get accidentally destroyed by a future commit?

vkrause:
> Would it make sense to have a test case to make sure this functionality doesn't get accidentally destroyed by a future commit?

Definitely. Added for the GTFS loader now, still looking for the best place to test the other parts.

include/nigiri/loader/gtfs/route.h
```diff
@@ -21,6 +21,8 @@ struct route {
   std::string long_name_;
   std::string desc_;
   clasz clasz_;
+  uint32_t color_;
```
felixguendling, line 24: I tried to use `std::uint32_t` everywhere (at least in new code like `nigiri`) instead of `uint32_t`. I also think it makes sense to introduce a strong type, so it cannot get confused with a regular integer (e.g. in parameter lists).

```cpp
using color_t = cista::strong<std::uint32_t, struct _color>;
```

include/nigiri/rt/frun.h
```diff
@@ -39,6 +39,8 @@ struct frun : public run {
     clasz get_clasz(event_type = event_type::kDep) const noexcept;
     clasz get_scheduled_clasz(event_type = event_type::kDep) const noexcept;
 
+    route_color route_color(event_type = event_type::kDep) const noexcept;
```
felixguendling, line 42:
```suggestion
    route_color get_route_color(event_type = event_type::kDep) const noexcept;
```

felixguendling: Our GCC version seems to not like that `struct route_color` and method `route_color` have the same name.

felixguendling:
> > Would it make sense to have a test case to make sure this functionality doesn't get accidentally destroyed by a future commit?
> 
> Definitely. Added for the GTFS loader now, still looking for the best place to test the other parts.

LGTM. I guess in MOTIS, only some kind of integration test with
- load GTFS
- check that output contains color attributes
is needed.

felixguendling:
Thank you! :-)

PR #68, opened 2024-02-18
Allow to set timeout for searching

jbruechert:


include/nigiri/routing/search.h
```diff
@@ -136,13 +138,27 @@ struct search {
...
       trace("start_time={}\n", search_interval_);
 
       search_interval();
 
       if (is_ontrip() || max_interval_reached() ||
-          n_results_in_interval() >= q_.min_connection_count_) {
+          n_results_in_interval() >= q_.min_connection_count_ ||
+          timeout_reached()) {
+        if (timeout_reached()) {
+          trace("  stopping search because timeout was reached\n");
```
felixguendling, line 159: I think it's easier to put this into the trace blow. Then it's sufficient to grep for `finished` to find the reason it's finished (besides being more compact code - less tracing code makes the actual logic more readable).

felixguendling:
Maybe a default argument would help to make the tests compile again.

include/nigiri/routing/search.h
```diff
@@ -136,17 +138,29 @@ struct search {
     state_.starts_.clear();
     add_start_labels(q_.start_time_, true);
 
+    auto processing_start_time = std::chrono::system_clock::now();
```
felixguendling, line 141:
```suggestion
    auto const processing_start_time = std::chrono::system_clock::now();
```

As it doesn't change.

include/nigiri/routing/search.h
```diff
@@ -136,17 +138,29 @@ struct search {
...
+    auto timeout_reached = [=, this]() -> bool {
```
felixguendling, line 142:
```suggestion
    auto const timeout_reached = [&]() -> bool {
```

As it doesn't change.

Capturing everything by `&` is usually the right thing to do. The compiler eliminates everything that's not used in the lambda. So no worries about copying a bunch of pointers around.

felixguendling:
Sorry, didn't see some `const` and the capture list in the first run. This should be the last change.

felixguendling:
I'm trying to upgrade `libfmt`.

felixguendling:
If you want to be fast, you can add the compiler definition `_SILENCE_ALL_MS_EXT_DEPRECATION_WARNINGS`. Otherwise, I am trying to update fmt in #70.

PR #69, opened 2024-02-20
Clang 17 compatibility

vkrause:
See individual commits for details, mainly fixing new warnings that break the build due to -Werror being set here.

That however only fixes half of the overall problem, `date` still needs a solution. That is unlikely to get upstream fixes for C++20 compatibility it seems (as it itself has become part of C++20 in slightly modified form), but with that new API appearing in libc++ we end up with a few clashes and/or ambiguities.

### DIFF
--- a/include/nigiri/loader/hrd/service/service.h
+++ b/include/nigiri/loader/hrd/service/service.h
@@ -41,8 +41,6 @@ struct specification {
 
 struct service {
   static const constexpr auto kTimeNotSet = -1;  // NOLINT
-  static category unknown_catergory;
-  static provider unknown_provider;
 
   struct event {
     int time_;
--- a/include/nigiri/types.h
+++ b/include/nigiri/types.h
@@ -207,19 +207,19 @@ using duration_t = i16_minutes;
 using unixtime_t = std::chrono::sys_time<i32_minutes>;
 using local_time = date::local_time<i32_minutes>;
 
-constexpr u8_minutes operator"" _i8_minutes(unsigned long long n) {
+constexpr u8_minutes operator""_i8_minutes(unsigned long long n) {
   return duration_t{n};
 }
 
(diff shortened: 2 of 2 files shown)


vkrause:
I don't think those changes are the problem here, it's mainly clashing ostream operator<< overloads it seems, upstream also has an open issues about that (https://github.com/HowardHinnant/date/issues/773).

felixguendling:
I was thinking about how to upgrade to the `std::` version of the `date` library. However, looking at the [compiler support](https://en.cppreference.com/w/cpp/compiler_support) it seems that the support is still "partial" which is not something I really want to experiment with - especially not if we can just stick to the `date` library which is really solid. Since we have our own fork of the `date` library anyway, we might as well fix the problems there.

A similar problem exists with the `fmt` library: moving to the `std::` version is not really worth it. The `fmt` library also gets performance improvements and new features here and there. So the `std::` version doesn't seem to be superior.

vkrause:
https://github.com/motis-project/date/pull/2 addresses the `date` problem.

oviano:
It's definitely a challenge trying to use this excellent library with C++20 when targeting various platforms:

Android - doesn't have zones, formatting or writing of durations and time points to streams
iOS/macOS - no zones, but does have formatting and streaming
Linux - has streaming
Windows - has everything

Obviously the above is specific to the versions of compilers I am using, but generally the most recent official.

felixguendling:
I guess I lack some context.

Regrading targets: currently we support MacOS (AppleClang), Windows (MSVC) and Linux (Clang 17 + GCC 13 for amd64 + aarch64) which is also tested via CI both in MOTIS as well as here. More targets (such as Android / iOS) are definitely possible (probably with slight changes here and there) but currently there's no use case so nobody invested the time.

PR #71, opened 2024-02-26
Classify unknown HVT route types based on the corresponding value range

vkrause:
Such types are for example in use in the Norwegian Entur feed for air and water transport, and were previously classified as "other".


felixguendling:
Thank you! :-)

PR #73, opened 2024-02-27
Significantly speed up applying GTFS-RT updates

vkrause:
With the national feeds from CH, NL and NO the time to process the updates goes down from about a minute to a second here. The cost of this was dominated by vecvec::push_back in rt_timetable::add_rt_transport.

Before:
gtfs-rt-update-before

After:
gtfs-rt-update-after


felixguendling:
To prevent bad memory accesses, I think it would make sense to call `operator[](index)` as a replacement for the `resize` calls. The code (routing, etc.) accesses the `rt_timetable` as `const`, so the `const` member function is called:

https://github.com/felixguendling/cista/blob/a46b7d806b6766175af2da3a45c4cc3f78a2d2c1/include/cista/containers/mutable_fws_multimap.h#L387-L397

This has an `assert` for debug mode but will crash in release mode. The rest of the code assumes that the vector has `n_locations` entries. As probably not all locations are touched during the real-time update, the update needs to call `operator[](n_locations)` instead of `resize` to prevent invalid memory accesses in routing where the non-resizing `const` member is used.

It probably makes sense to give `mutable_fws_multimap` the same interface as `vecvec` (so not mirroring a `std::multimap` but rather a `std::vector<std::vector<T>>`).

PR #81, opened 2024-04-29
Allow csv unescaping for station names

jbruechert:
This will only start doing something once https://github.com/motis-project/utl/pull/18 is merged

It is the minimum amount of changes to fix the issue  I encountered, but should probably be done for more fields.


felixguendling:
Wouldn't it be sufficient to have `csv_range` not create a new `T` every time `read_row` is called?
https://github.com/motis-project/utl/blob/4c1503afe58e209977d9a1e3db6a6b271a50c521/include/utl/parser/csv_range.h#L94

If it would have a `T` member variable that it can reuse every time `read_row` is called, allocations would be very limited.

Having global state is something I would try to avoid as it can introduce unexpected problems in with multi-threading, etc.

felixguendling:
True. You're right.

I think I'll try to change the procedure a bit so data is directly stored to the `timetable` struct.

felixguendling:
I think another option would be to use `cista::raw::basic_string` as it act as both: a non-owning `string_view` or a owning `string`. So in 99.9% of the time it would act as `string_view` but in case an allocation is needed, it becomes a `string` and allocates. Unfortunately, it doesn't have a `std::string` compatible interface (like `erase`).

felixguendling:
Next step would be to update `utl` and use `cista::raw::generic_string`, right? Or is anything else blocking? Just want to make sure I didn't miss any other PR where you're blocked waiting for my review.

felixguendling:
If you work directly on the nigiri repo (not in deps, but nigiri = root), pkg should not interfere.

pkg in general only checks out a version if `current_ref != target_ref` where `target_ref` is the latest commit referenced over the whole tree. So if you set this to your current commit, there should not be any resets.

felixguendling:
Otherwise, you can just replace `include(cmake/pkg.cmake)` with `add_subdirectory(deps)` to disable the check.

felixguendling:
My mistake. I missed to return `*this` in the assignment operators of `generic_string`. I updated `utl` to have a mutable `csv_col::operator*` that returns a mutable reference which should enable `std::move` to work. In case it works, could you please benchmark again so we make sure there's no performance degradation for importing data?

include/nigiri/location.h
```diff
@@ -26,9 +35,10 @@ struct location {
            timezone_idx_t,
            duration_t transfer_time,
            it_range<vector<location_idx_t>::const_iterator> equivalences);
+
   location_idx_t l_{location_idx_t::invalid()};
   std::string_view id_;
-  std::string_view name_;
+  cista::raw::generic_string name_;
```
felixguendling, line 41: location is supposed to be a "view" on data, not own the data, so it should stay a std::string_view as it was before

felixguendling (changes requested):

felixguendling:
I think it was good with the version we had before. The latest change with making `location::name_` a `generic_string` didn't work and is also not necessary as location is always only a (volatile/temporary) view on data.

felixguendling:
With this change address sanitizer doesn't complain anymore:
https://github.com/felixguendling/nigiri/commit/07f6ec71395ead56ce42cec924ed93b00d70755b

felixguendling:
What's the performance impact of the new solution?

felixguendling (approved):

felixguendling:
Looks good! :-)

PR #82, opened 2024-04-30
Importer

mority:
Executable that loads a nigiri timetable from GTFS or HRD data sources and creates a binary dump of it


importer/main.cc
```diff
@@ -0,0 +1,117 @@
...
+  if (output_file.has_filename() && !output_file.has_parent_path()) {
+    if (is_directory(input_path)) {
+      output_path = input_path;
+    } else {
+      output_path = input_path.parent_path();
+    }
+    output_path /= output_file;
+  } else if (is_directory(output_file)) {
+    output_path /= "tt.bin";
+  }
```
felixguendling, line 107: Maybe use default values to make the code more readable and usage a bit easier?

```cpp
auto output_file = std::filesystem::path{};
desc.add_options()
  ("out", bpo::value<std::filesystem::path>()->default_value(output_file), 
   "the name of the output file");
```

felixguendling (changes requested):

importer/main.cc
felixguendling: default values could be

- start_date=TODAY
- end_date=TODAY+365

felixguendling: Probably easier to use `num_days` as in MOTIS

PR #83, opened 2024-05-02
Bitvec

mority:
replaces std::vector<bool> with cista's bitvec to make use of its for_each_set_bit function


felixguendling:
Can you please post the benchmark results (just for reference)?

PR #105, opened 2024-06-04
Shortest footpath

mority:
Adds a postprocessing step after reconstruction. 

- Tries to replace the start station with a station that has a departure of the same transport as the first transport leg of the journey but is reachable from the START coordinate by a shorter footpath
- Tries to replace the end station with a station that has an arrvial of the same transport as the last transport leg of the journey but reaches the END coordinate with a shorter footpath
- For each transfer tries to find a shorter footpath that allows to transfer between the two transports involved

### DIFF
--- a/src/routing/raptor/optimize_footpaths.cc
+++ b/src/routing/raptor/optimize_footpaths.cc
@@ -1,21 +1,52 @@
 #include "nigiri/routing/raptor/reconstruct.h"
 
+#include "nigiri/routing/for_each_meta.h"
 #include "nigiri/routing/journey.h"
+#include "nigiri/rt/frun.h"
+#include "nigiri/timetable.h"
 
 namespace nigiri::routing {
 
-void optimize_footpaths(timetable const& tt,
-                        rt_timetable const* rtt,
-                        query const& q,
-                        raptor_state const& raptor_state,
-                        journey& j,
-                        date::sys_days const base,
-                        day_idx_t const base_day_idx) {
-  // optimize start
+void optimize_start(timetable const& tt, query const& q, journey& j) {
+  if (j.legs_.size() <= 1 || !holds_alternative<offset>(j.legs_[0].uses_) ||
+      !holds_alternative<journey::run_enter_exit>(j.legs_[1].uses_)) {
+    return;
+  }
+  auto& offset_leg = j.legs_[0];
+  auto& transport_leg = j.legs_[1];
+  auto& ree = get<journey::run_enter_exit>(transport_leg.uses_);
+  auto offset_dur_best = get<offset>(offset_leg.uses_).duration();
+  for (auto const& o : q.start_) {
+    if (offset_dur_best <= o.duration()) {
+      continue;
+    }
+    auto const stop_seq =
+        tt.route_location_seq_[tt.transport_route_[ree.r_.t_.t_idx_]];
+    for (auto stop_idx = stop_idx_t{0U}; stop_idx != ree.stop_range_.to_;
+         ++stop_idx) {
+      auto stp = stop{stop_seq[stop_idx]};
+      if (!stp.in_allowed() || !matches(tt, location_match_mode::kEquivalent,
+                                        o.target(), stp.location_idx())) {
+        continue;
+      }
+      auto const dep = tt.event_time(ree.r_.t_, stop_idx, event_type::kDep);
+      auto const o_start = dep - o.duration();
+      if (offset_leg.dep_time_ <= o_start) {
+        offset_leg.to_ = stp.location_idx();
+        offset_leg.dep_time_ = o_start;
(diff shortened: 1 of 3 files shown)


src/routing/raptor/optimize_footpaths.cc
```diff
@@ -0,0 +1,234 @@
...
+        !holds_alternative<journey::run_enter_exit>(j.legs_[i + 2].uses_) ||
+        matches(tt, location_match_mode::kEquivalent, leg.to_,
+                leg_transfer.from_)) {
+      continue;
+    }
+    auto fp_dur_best = get<footpath>(leg_footpath.uses_).duration();
+    auto& ree = get<journey::run_enter_exit>(leg.uses_);
+    auto& ree_transfer = get<journey::run_enter_exit>(leg_transfer.uses_);
+
+    auto const iterate_stops = [&](auto&& stop_seq) {
```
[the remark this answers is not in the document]
felixguendling: Maybe even `frun` can be used?

src/routing/raptor/optimize_footpaths.cc
```diff
@@ -0,0 +1,234 @@
...
+
+#include "nigiri/routing/for_each_meta.h"
+#include "nigiri/routing/journey.h"
+#include "nigiri/rt/frun.h"
+#include "nigiri/rt/rt_timetable.h"
+#include "nigiri/timetable.h"
+
+namespace nigiri::routing {
+
+void optimize_start(timetable const& tt,
```
felixguendling, line 11: Does this consider that `start` in backward search means the last leg of the journey (-> arrival times are relevant)?

src/routing/raptor/optimize_footpaths.cc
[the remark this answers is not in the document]
mority: > why a lambda? Can't stop_seq be a ordinary variable that's assigned
> 
> ```c++
> (rtt != nullptr && rtt->resolve_rt(ree.r_.t_) != rt_transport_idx_t::invalid())
>     ? rtt->rt_transport_location_seq_[rtt->resolve_rt(ree.r_.t_)]
>     : tt.route_location_seq_[tt.transport_route_[ree.r_.t_.t_idx_]]
> ```
I tried to code it as seen below initially.
```
auto const stop_seq =
    (rtt != nullptr &&
     rtt->resolve_rt(ree.r_.t_) != rt_transport_idx_t::invalid())
        ? rtt->rt_transport_location_seq_[rtt->resolve_rt(ree.r_.t_)]
        : tt.route_location_seq_[tt.transport_route_[ree.r_.t_.t_idx_]];
```
However, the types mismatch due to one using rt_transport_idx and the other route_idx. Even though, they both are vecvec with stop::value_type as their second template type. Maybe there is a way to unpack it further to get rid of the mismatching first type.
```
error: incompatible operand types ('const_bucket' (aka 'cista::basic_vecvec<cista::strong<unsigned int, nigiri::_rt_transport_idx>, cista::basic_vector<unsigned int, cista::raw::ptr>, cista::basic_vector<unsigned int, cista::raw::ptr>>::const_bucket') and 'const_bucket' (aka 'cista::basic_vecvec<cista::strong<unsigned int, nigiri::_route_idx>, cista::basic_vector<unsigned int, cista::raw::ptr>, cista::basic_vector<unsigned int, cista::raw::ptr>>::const_bucket'))
   31 |             ? rtt->rt_transport_location_seq_[rtt->resolve_rt(ree.r_.t_)]
      |             ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   32 |             : tt.route_location_seq_[tt.transport_route_[ree.r_.t_.t_idx_]];
      |               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Regardless, using frun I could remove the lambdas and simply iterate over the stops of the frun. I pushed a new version doing just that.

src/routing/raptor/optimize_footpaths.cc
mority: Good point. So, the legs of the journey are in the order in which a real-world traveller would make the journey. However, if a backward search was used to answer the query, the offsets at the real-world start are in `q.destination_` of the query, right?

test/routing/test_data.cc
```diff
@@ -0,0 +1,100 @@
+#include "./test_data.h"
```
felixguendling, line 1: I think this style has better readability

https://github.com/motis-project/nigiri/blob/master/test/rt/ripta_test.cc

felixguendling: as long as the test case is only used in one file, it's sufficient to have it only there (to prevent having to write a `.h` file and a `.cc` file instead of just having the test inline).

PR #108, opened 2024-06-12
Bike transport

pablohoch:
Bike transport support:
- GTFS: `bikes_allowed` field in `trips.txt` is now parsed.
- For trips with `bikes_allowed = 0` or no bike information, a default value based on the route clasz is used (defaults can be specified in `loader_config`).
- Added a new `require_bikes_allowed_` flag to `routing::query`. If set to false (the default value), returns the same results as before. If set to true, only trips where bikes are allowed can be used.

Additional changes:
- Changes to footpaths in reconstructed journeys:
  - A 0 minute footpath is no longer added at the end (forward search) / beginning (backward search) of a journey
  - Footpaths now always start at the arrival time of the previous leg (previously, they ended at the departure time of the next leg when searching backwards)
- `journey.print()` now always prints the departure - arrival interval at the top (previously: arrival - departure if using backward search)


include/nigiri/loader/gtfs/trip.h
```diff
@@ -23,6 +23,9 @@ using gtfs_trip_idx_t = cista::strong<std::uint32_t, struct _gtfs_trip_idx>;
 
 struct trip_data;
 
+static auto const kSingleTripBikesAllowed = bitvec("1");
+static auto const kSingleTripBikesNotAllowed = bitvec("0");
```
felixguendling, line 27:
```suggestion
static auto const kSingleTripBikesAllowed = bitvec{"1"};
static auto const kSingleTripBikesNotAllowed = bitvec{"0"};
```

felixguendling (approved):

felixguendling:
Very nice! :+1:

PR #113, opened 2024-06-19
Add transfer time settings for routing

pablohoch:
Adds two new options to routing queries:
- Minimum transfer time: All transfer times and footpath durations shorter than the minimum transfer time are set to the minimum transfer time
- Transfer time factor: All transfer times and footpath durations are multiplied by this factor (applied before the minimum transfer time check)

This version uses the original lower bounds. Alternative version with adjusted lower bounds: #114


pablohoch:
Benchmark master vs. this branch with default transfer time settings -> https://github.com/motis-project/nigiri/pull/122#issuecomment-2286353857

PR #127, opened 2024-08-29
Add support for shapes.txt

MichaelKutzner:
Add support for reading `shapes.txt` to a memory map. This data can later be used to request the shape of a trip.


MichaelKutzner:
As I noticed, MSVC doesn't like trailing return types. I'll update the code, but I'll take until next week.

felixguendling (approved):

PR #130, opened 2024-09-06
Allow optional creation mode

MichaelKutzner:
Making the creation mode optional allows clients to load stored data without duplicating this function. This is especially useful if import and usage can run independently.
For instance, it could be applied to motis-project/motis#537 immediately to remove duplicated code.

### DIFF
--- a/include/nigiri/shape.h
+++ b/include/nigiri/shape.h
@@ -9,7 +9,7 @@
 
 namespace nigiri {
 struct timetable;
-}  // namespace nigiri
+}
 
 namespace nigiri {
 
--- a/include/nigiri/shape.h
+++ b/include/nigiri/shape.h
@@ -9,11 +9,13 @@
 
 namespace nigiri {
 struct timetable;
-}
+}  // namespace nigiri
 
 namespace nigiri {
 
-shapes_storage_t create_shapes_storage(std::filesystem::path const&);
+shapes_storage_t create_shapes_storage(
+    std::filesystem::path const&,
+    cista::mmap::protection = cista::mmap::protection::WRITE);
 
 std::span<geo::latlng const> get_shape(timetable const&,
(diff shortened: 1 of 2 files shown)


include/nigiri/shape.h
```diff
@@ -9,11 +9,13 @@
 
 namespace nigiri {
 struct timetable;
-}
+}  // namespace nigiri
```
felixguendling, line 12:
```suggestion
}
```

src/shape.cc
```diff
@@ -6,8 +6,8 @@
 
 namespace nigiri {
 
-shapes_storage_t create_shapes_storage(std::filesystem::path const& path) {
-  constexpr auto const kMode = cista::mmap::protection::WRITE;
+shapes_storage_t create_shapes_storage(std::filesystem::path const& path,
+                                       cista::mmap::protection const kMode) {
   return {
       cista::basic_mmap_vec<geo::latlng, std::uint64_t>{cista::mmap{
           fmt::format("{}_data.bin", path.generic_string()).c_str(), kMode}},
```
felixguendling, line 13:
```suggestion
shapes_storage_t create_shapes_storage(std::filesystem::path const& path,
                                       cista::mmap::protection const mode) {
  return {
      cista::basic_mmap_vec<geo::latlng, std::uint64_t>{cista::mmap{
          fmt::format("{}_data.bin", path.generic_string()).c_str(), mode}},
```

felixguendling: `kCamelCase` is a compile time constant

felixguendling (changes requested):

src/shape.cc
MichaelKutzner: Thank you for explaining. I didn't know about this convention.

PR #134, opened 2024-09-13
Reached à la Trip-based

mority:
For each route $r$ maintain a Pareto set of tuples $(t, s, k) \in T \times \mathbb{N} \times \mathbb{N}$ where $T$ is the set of sorted transport indices, i.e., the difference in the day index is stored in the most significant bits of the transport index, $s$ is the stop index and $k$ is the round of the RAPTOR algorithm.


mority:
Some VIA test cases do not work yet

PR #136, opened 2024-09-16
Calculate stop to subshape mapping during import

MichaelKutzner:
This extends the shape support, implemented with #127.
Clients may now request the shape of a trip segment. These will be calculated during import, to reduce runtime costs. The corresponding mapping will be stored to a memory map.

As the calculation will take some time, especially for frequent trips using the same stops and shape, results will be cached during import.


MichaelKutzner:
Marked as draft again, as using `shape_dist_traveled` is almost finished.

src/loader/gtfs/shape_prepare.cc
```diff
@@ -0,0 +1,169 @@
...
+      };
+  auto shape_offsets_cache =
+      hash_map<std::pair<shape_idx_t, stop_seq_t const*>, shape_offset_idx_t,
+               decltype(key_hash), decltype(key_compare)>{};
+  for (auto const& trip : trips) {
+    progress_tracker->increment();
+    auto const trip_index = trip.trip_idx_;
+    auto const shape_index = trip.shape_idx_;
+    auto const shape_distances = shapes_distances.find(shape_index);
+    if (shape_distances != std::end(shapes_distances) &&
```
felixguendling, line 143:
```suggestion
    if (shape_distances != end(shapes_distances) &&
```

src/loader/gtfs/shape_prepare.cc
```diff
@@ -0,0 +1,169 @@
...
+      offsets[0] = shape_offset_t{0};
+    } else if (index == stops.size() - 1U) {
+      offsets[index] = shape_offset_t{shape.size() - 1U};
+    } else {
+      auto const location =
+          tt.locations_.get(stop(location_index).location_idx());
+      offsets[index] = offset += get_closest(
+          location.pos_, shape.subspan(static_cast<std::size_t>(offset.v_)));
+    }
+    ++index;
```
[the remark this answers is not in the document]
MichaelKutzner: Good hint. I missed that. The code has been updated.

src/loader/gtfs/shape_prepare.cc
```diff
@@ -0,0 +1,169 @@
...
+  if (shape.empty()) {
+    return {};
+  }
+  auto offsets = std::vector<shape_offset_t>(stops.size());
+  auto offset = shape_offset_t{0};
+
+  auto index = 0U;
+  for (auto const& location_index : stops) {
+    if (index == 0U) {
+      offsets[0] = shape_offset_t{0};
```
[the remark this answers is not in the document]
MichaelKutzner: I've updated all initialization values, I think.

src/loader/gtfs/shape_prepare.cc
```diff
@@ -0,0 +1,169 @@
...
+    } else if (index == stops.size() - 1U) {
+      offsets[index] = shape_offset_t{shape.size() - 1U};
+    } else {
+      auto const location =
+          tt.locations_.get(stop(location_index).location_idx());
+      offsets[index] = offset += get_closest(
+          location.pos_, shape.subspan(static_cast<std::size_t>(offset.v_)));
```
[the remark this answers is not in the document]
MichaelKutzner: Thank you for that suggestion. I've updated this part to be more readable.

test/rt/rt_block_id_test.cc
```diff
@@ -130,4 +137,64 @@ TEST(rt, rt_block_id_test) {
...
+
+      fr.for_each_shape_point(nullptr, run_ee.stop_range_,
+                              [&leg_shape](geo::latlng const& point) {
+                                leg_shape.push_back(point);
+                              });
+
+      auto const expected_shape =
+          std::vector{geo::latlng{2.0F, 3.0F}, geo::latlng{4.0F, 5.0F},
+                      geo::latlng{6.0F, 7.0F}, geo::latlng{8.0F, 9.0F}};
+      EXPECT_EQ(expected_shape, leg_shape);
```
felixguendling, line 170: polyline

include/nigiri/loader/gtfs/shape.h
```diff
@@ -11,8 +16,21 @@ struct shape_state {
...
+  shape_idx_t index_offset_;
+};
+
+shape_loader_state parse_shapes(std::string_view const, shapes_storage&);
 
-shape_id_map_t parse_shapes(std::string_view const, shapes_storage_t&);
+template <typename DoubleRange>
+  requires std::ranges::range<DoubleRange> &&
+           std::is_same_v<std::ranges::range_value_t<DoubleRange>, double>
+inline bool valid_distances(DoubleRange distances) {
```
felixguendling, line 31:
```suggestion
bool valid_distances(DoubleRange distances) {
```

Templates do not need `inline` and usually the compiler does a better job at deciding what should be really inlined. So `inline` is mostly a instruction to the linker, which is not needed in this case (as it's a template).

MichaelKutzner: Thank you for the reminder.
Due to some changes, the function has been removed entirely.

felixguendling:
Amazing job!! :star_struck: :rocket: :tada:

PR #139, opened 2024-09-24
Extend ABI to expose journeys, footpaths, cancellations

traines-source:
This finally addresses most of the open todos from #66, among other things:
* handle RT track changes, in_allowed/out_allowed
* expose footpaths
* expose querying/journeys

There is one change I made that has an impact outside of the ABI: For absolute-time-based GTFSRTs (`ev.has_time()`), the time-travel-corrected time will be propagated in [L119](https://github.com/traines-source/nigiri/blob/aa4c837b46b62f69d94295d9cf92d2c2ad13bfa2/src/rt/gtfsrt_update.cc#L119) just like for delay-based GTFSRTs in [L94](https://github.com/traines-source/nigiri/blob/aa4c837b46b62f69d94295d9cf92d2c2ad13bfa2/src/rt/gtfsrt_update.cc#L94). I think this was a bug.

### DIFF
--- a/src/rt/gtfsrt_update.cc
+++ b/src/rt/gtfsrt_update.cc
@@ -85,11 +85,12 @@ delay_propagation update_delay(timetable const& tt,
                                duration_t const delay,
                                std::optional<unixtime_t> const min) {
   auto const static_time = tt.event_time(r.t_, stop_idx, ev_type);
-  rtt.update_time(r.rt_, stop_idx, ev_type,
-                  min.has_value() ? std::max(*min, static_time + delay)
-                                  : static_time + delay);
+  auto const lower_bounded_new_time = min.has_value()
+                                          ? std::max(*min, static_time + delay)
+                                          : static_time + delay;
+  rtt.update_time(r.rt_, stop_idx, ev_type, lower_bounded_new_time);
   dispatch_event_change(rtt, r, stop_idx, ev_type, location_idx_t{0}, -1,
-                        delay);
+                        lower_bounded_new_time - static_time);
   return {rtt.unix_event_time(r.rt_, stop_idx, ev_type), delay};
 }
 
@@ -110,12 +111,12 @@ delay_propagation update_event(timetable const& tt,
     auto const new_time =
         unixtime_t{std::chrono::duration_cast<unixtime_t::duration>(
             std::chrono::seconds{ev.time()})};
-    rtt.update_time(
-        r.rt_, stop_idx, ev_type,
-        pred_time.has_value() ? std::max(*pred_time, new_time) : new_time);
+    auto const lower_bounded_new_time =
+        pred_time.has_value() ? std::max(*pred_time, new_time) : new_time;
+    rtt.update_time(r.rt_, stop_idx, ev_type, lower_bounded_new_time);
     dispatch_event_change(rtt, r, stop_idx, ev_type, location_idx_t{0}, -1,
-                          new_time - static_time);
-    return {new_time, new_time - static_time};
+                          lower_bounded_new_time - static_time);
+    return {lower_bounded_new_time, new_time - static_time};
   }
 }
 
--- a/src/rt/gtfsrt_update.cc
+++ b/src/rt/gtfsrt_update.cc
@@ -63,6 +63,20 @@ std::string remove_nl(std::string s) {
   return s;
 }
 
+void dispatch_event_change(rt_timetable& rtt,
+                           run const& r,
+                           stop_idx_t const stop_idx,
+                           event_type const ev_type,
+                           location_idx_t const location_idx,
+                           int16_t const in_out_allowed,
+                           duration_t const delay) {
+  if ((ev_type == event_type::kArr && stop_idx != r.stop_range_.from_) ||
+      (ev_type == event_type::kDep && stop_idx != r.stop_range_.to_ - 1)) {
+    rtt.dispatch_event_change(r.t_, stop_idx, ev_type, location_idx,
+                              in_out_allowed, delay);
+  }
+}
+
 delay_propagation update_delay(timetable const& tt,
                                rt_timetable& rtt,
                                run const& r,
@@ -74,7 +88,8 @@ delay_propagation update_delay(timetable const& tt,
   rtt.update_time(r.rt_, stop_idx, ev_type,
                   min.has_value() ? std::max(*min, static_time + delay)
                                   : static_time + delay);
-  rtt.dispatch_event_change(r.t_, stop_idx, ev_type, delay, false);
+  dispatch_event_change(rtt, r, stop_idx, ev_type, location_idx_t{0}, -1,
+                        delay);
   return {rtt.unix_event_time(r.rt_, stop_idx, ev_type), delay};
 }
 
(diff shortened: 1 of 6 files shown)


felixguendling:
Regarding formatting: We switched to Clang-18 (`clang-format-18`).

felixguendling:
@mority can you please check the GTFS RT part with the proposed propagation bugfix? Thank you :)

traines-source:
Regarding formatting: I just forgot to format this file, and the `.clang-format` config, which my editor should use, seems to be fine even if it has not been updated in the past two years?

More details regarding propagation: As far as I understood this was meant to avoid vehicles travelling back in time due to missing or wrong realtime updates, e.g. having a delay of one hour at one stop and arriving on time at the next stop which is only a minute away. I think I even encountered this in the wild back in the day (with the RT feed from gtfs.de which uses absolute timestamps), but I'm not sure anymore, would have to check.

src/rt/gtfsrt_update.cc
```diff
@@ -71,10 +85,12 @@ delay_propagation update_delay(timetable const& tt,
...
   auto const static_time = tt.event_time(r.t_, stop_idx, ev_type);
-  rtt.update_time(r.rt_, stop_idx, ev_type,
-                  min.has_value() ? std::max(*min, static_time + delay)
-                                  : static_time + delay);
-  rtt.dispatch_event_change(r.t_, stop_idx, ev_type, delay, false);
+  auto const lower_bounded_new_time = min.has_value()
+                                          ? std::max(*min, static_time + delay)
+                                          : static_time + delay;
+  rtt.update_time(r.rt_, stop_idx, ev_type, lower_bounded_new_time);
+  dispatch_event_change(rtt, r, stop_idx, ev_type, location_idx_t{0}, -1,
```
mority, line 92: `location_idx_t::invalid()` could be used to signal that this is a dummy value.

The sentinel value `-1` to signal no change regarding `in_out_allowed` is only specified in `abi.h` which is not included by this file. For a future reader it is not immediately clear what `-1` means. Maybe we could use an `enum enter_exit_update` or something like that to make it more readable what the value means that is passed here.

src/rt/gtfsrt_update.cc
```diff
@@ -95,12 +111,12 @@ delay_propagation update_event(timetable const& tt,
...
-    rtt.update_time(
-        r.rt_, stop_idx, ev_type,
-        pred_time.has_value() ? std::max(*pred_time, new_time) : new_time);
-    rtt.dispatch_event_change(r.t_, stop_idx, ev_type, new_time - static_time,
-                              false);
-    return {new_time, new_time - static_time};
+    auto const lower_bounded_new_time =
+        pred_time.has_value() ? std::max(*pred_time, new_time) : new_time;
+    rtt.update_time(r.rt_, stop_idx, ev_type, lower_bounded_new_time);
+    dispatch_event_change(rtt, r, stop_idx, ev_type, location_idx_t{0}, -1,
```
mority, line 117: same as line 92

src/rt/gtfsrt_update.cc
```diff
@@ -113,6 +129,13 @@ void cancel_run(timetable const&, rt_timetable& rtt, run& r) {
     rtt.bitfields_.emplace_back(bf).set(to_idx(r.t_.day_), false);
     rtt.transport_traffic_days_[r.t_.t_idx_] =
         bitfield_idx_t{rtt.bitfields_.size() - 1U};
+
+    for (auto i = r.stop_range_.from_; i != r.stop_range_.to_; ++i) {
+      dispatch_event_change(rtt, r, i, event_type::kArr, location_idx_t{0}, 0,
```
mority, line 134: same as line 92

src/rt/gtfsrt_update.cc
```diff
@@ -113,6 +129,13 @@ void cancel_run(timetable const&, rt_timetable& rtt, run& r) {
...
+                            duration_t{0});
+      dispatch_event_change(rtt, r, i, event_type::kDep, location_idx_t{0}, 0,
```
mority, line 136: same as line 92

src/rt/gtfsrt_update.cc
```diff
@@ -113,6 +129,13 @@ void cancel_run(timetable const&, rt_timetable& rtt, run& r) {
     rtt.bitfields_.emplace_back(bf).set(to_idx(r.t_.day_), false);
     rtt.transport_traffic_days_[r.t_.t_idx_] =
         bitfield_idx_t{rtt.bitfields_.size() - 1U};
+
+    for (auto i = r.stop_range_.from_; i != r.stop_range_.to_; ++i) {
+      dispatch_event_change(rtt, r, i, event_type::kArr, location_idx_t{0}, 0,
+                            duration_t{0});
```
mority, line 135: parameter could default to `0` if time does not change

src/rt/gtfsrt_update.cc
```diff
@@ -163,6 +186,10 @@ void update_run(
         // Cancel skipped stops (in_allowed = out_allowed = false).
         stp =
             stop{stop{stp}.location_idx(), false, false, false, false}.value();
+        dispatch_event_change(rtt, r, stop_idx, event_type::kArr,
```
mority, line 189: Overall, it seems as if `dispatch_event_changes` has to use a lot of placeholder arguments that do not give information to the reader on what is happening. We should think about using more descriptive function names, e.g. `dispatch_cancelation(rtt, r, stop_idx, event_type::kArr)` could then internally pass the correct null values. The same could be done for the other use cases of `dispatch_event_change(...)`

src/rt/vdv/vdv_update.cc
```diff
@@ -332,7 +332,8 @@ void update_event(rt_timetable& rtt,
             et == event_type::kArr ? "ARR" : "DEP", rs.scheduled_time(et),
             delay.count() >= 0 ? "+" : "", delay.count());
   rtt.update_time(rs.fr_->rt_, rs.stop_idx_, et, new_time);
-  rtt.dispatch_event_change(rs.fr_->t_, rs.stop_idx_, et, delay, false);
+  rtt.dispatch_event_change(rs.fr_->t_, rs.stop_idx_, et, location_idx_t{0},
```
mority, line 335: see comments on `gtfsrt_update.cc`

src/rt/vdv/vdv_update.cc
```diff
@@ -372,7 +373,8 @@ void updater::update_run(rt_timetable& rtt,
               et == event_type::kArr ? "ARR" : "DEP", rs.scheduled_time(et),
               delay->count() >= 0 ? "+" : "", delay->count());
     rtt.update_time(fr.rt_, rs.stop_idx_, et, rs.scheduled_time(et) + *delay);
-    rtt.dispatch_event_change(fr.t_, rs.stop_idx_, et, *delay, false);
+    rtt.dispatch_event_change(fr.t_, rs.stop_idx_, et, location_idx_t{0}, false,
```
mority, line 376: see comments on `gtfsrt_update.cc`

traines-source:
Sorry for the wait, good points. I have refactored accordingly, but instead of an enum I have used `std::optional` on the C++ side and an additional bool on the C side, because the delay actually also needs to be ignored/undefined for events where stop metadata (track or in/out_allowed) is changed.

PR #149, opened 2024-10-24
storage shapes in paged data structures for improved performance on unordered files

felixguendling:


felixguendling:
image


currently produces bad shapes

PR #150, opened 2024-11-13
Change file name for shape point indices

MichaelKutzner:
Changing a filename, that is probably misleading.


felixguendling:
don't forget to increment the nigiri binary version in MOTIS for this change

MichaelKutzner:
Thank you for the reminder. I just created the corresponding pull request, so it isn't missed.

PR #153, opened 2024-11-15
Add setting for disposable time between transfers

MichaelKutzner:
This adds an option to increase the time between transfers by a fixed duration. This setting can be used, if passengers want some time for orientation or to compensate small delays.

### DIFF
--- a/include/nigiri/routing/search.h
+++ b/include/nigiri/routing/search.h
@@ -87,7 +87,7 @@ struct search {
 
     tts.factor_ = std::max(tts.factor_, 1.0F);
     if (tts.factor_ == 1.0F && tts.min_transfer_time_ == 0_minutes &&
-        tts.disposable_time_ == 0_minutes) {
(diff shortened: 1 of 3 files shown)


felixguendling:
please rename disposable -> additional

the word disposable seems not like a good fit
https://www.thesaurus.com/browse/disposable

MichaelKutzner:
~~Any suggestions for a better name?~~ _Thank you for the fast reply_
I'm also not sure, if the duration for the footpath should include this duration. Or if it should be split, e.g. `FOOTPATH (duration=X,  additional_time=Y)`

felixguendling:
I guess the user expects that (footpath time + additional time) < (departure time - arrival time)

At least I would expect that.

MichaelKutzner:
That was my intention, to make the footpath duration and additional time more obvious. But as I wrote, I'm not sure what format might be good to avoid possible confusion. I think, it might also depends on how the feature will be used.

PR #155, opened 2024-11-21
Add configurable maximum travel time

MichaelKutzner:
This adds a new query parameter `max_travel_time_` to limit results by maximum travel time.

Notice that this change also includes tests for `max_transfers_`, which serve a similar purpose. As these has been added to `test/routing/transfer_time_settings_test.cc`, the file will be renamed to `test/routing/transfer_travel_test.cc`.

### DIFF
--- a/include/nigiri/routing/query.h
+++ b/include/nigiri/routing/query.h
@@ -2,6 +2,7 @@
 
 #include <cinttypes>
 #include <limits>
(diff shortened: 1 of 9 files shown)


MichaelKutzner:
Closed, as this will not prevent calculating results, that will be discarded later on.

PR #158, opened 2024-12-05
Via search fixes

pablohoch:
Fixes the following bugs in the via search implementation that caused reconstruction errors:

- If a trip passing a via stop has in/out allowed = false at the via stop, reconstructing the journey failed.
- Queries where the last via stop is also a destination and has a non-zero stay duration were not handled properly. They are now handled as follows:
  - Intermodal destination (probably the only useful case): Stay duration at the last stop is added as waiting time between the arrival + start of the intermodal footpath. Transfer time at the last stop is ignored.
  - Station destination: Stay duration at the last stop is ignored (i.e. same result as if stay = 0 in the query). The stay duration can't be properly represented in the output journey in this case, but this doesn't seem like a useful case anyway.

### DIFF
--- a/include/nigiri/routing/raptor/raptor.h
+++ b/include/nigiri/routing/raptor/raptor.h
@@ -592,8 +592,8 @@ private:
           if (best_time == kInvalid) {
             return;
           }
-          auto const stay = Vias != 0 && is_via_[Vias - 1][i]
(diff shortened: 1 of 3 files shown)


felixguendling:
🎉

PR #159, opened 2024-12-11
Don't drop partially unsupported GTFS-RT messages

jbruechert:
Previously all messages containing unsupported fields like (alert or
vehicle) were ignored, even if they contained an understandable
trip_update.

Now all possible warnings are emitted, not just the first fatal one.

This commit also changes the debug output to show the names of the
relevant fields to make it easier to debug GTFS-RT feeds without looking
at the MOTIS source code.

### DIFF
--- a/src/rt/gtfsrt_update.cc
+++ b/src/rt/gtfsrt_update.cc
@@ -268,58 +268,49 @@ statistics gtfsrt_update_msg(timetable const& tt,
   span->SetAttribute("nigiri.gtfsrt.total_entities", msg.entity_size());
 
   for (auto const& entity : msg.entity()) {
-    bool skip = false;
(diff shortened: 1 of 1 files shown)


jbruechert:
Apparently the specification recommends against sending multiple types of update in one message (https://github.com/public-transport/transitous/pull/689#issuecomment-2535533503). 
Does it still make sense to be lenient on the consumer side and leave interpreting the spec strictly to the producers, or should I reduce this PR to the logging changes?

felixguendling:
As long as there's only this one feed that is not standard conformant (and it seems like it can be changed) I would prefer to stick to the standard. So logging a warning message if the sum is `!= 1U` would be sufficient IMO.

In case feeds we want to support but don't have any control over can be support with a small change, I would merge it.

PR #165, opened 2025-01-17
Adopting utl::log()

Lucas-C:


felixguendling:
I think we need an option in `utl` to enable the docs.
`UTL_BUILD_DOCS` with default value `OFF`.

And then wrap this in `if (UTL_BUILD_DOCS)`:
https://github.com/motis-project/utl/blob/69e700e03ae3d385304d754cf57877528bc591ef/CMakeLists.txt#L33

felixguendling:
Seeing the code, maybe it makes sense to rename to something with `log` in the name (e.g. `utl::log_err`)? At least `utl::error` could also be an exception class derived from `std::exception`.

Lucas-C:
> I think we need an option in `utl` to enable the docs. `UTL_BUILD_DOCS` with default value `OFF`.
> 
> And then wrap this in `if (UTL_BUILD_DOCS)`: [motis-project/utl@`69e700e`/CMakeLists.txt#L33](https://github.com/motis-project/utl/blob/69e700e03ae3d385304d754cf57877528bc591ef/CMakeLists.txt#L33)

Do all the MOTIS repos builds rely on the `CMakeLists.txt` in the `utl` module?

I'm not a big CMake expert, so I'm not sure how best to share the doc-building logic between MOTIS repositories.
I think the following could be shared from `motis-project/utl`:
* `docs/CMakeLists.txt`
* docs/Doxyfile (except from the module name)
* `docs/conf.py` (except from the module name)
* `docs/motis-logo.svg`
* `docs/requirements.txt`
* the `doc` job from `.github/workflows/unix.yml`

Lucas-C:
> Seeing the code, maybe it makes sense to rename to something with `log` in the name (e.g. `utl::log_err`)? At least `utl::error` could also be an exception class derived from `std::exception`.

Yes, I agree, I think it would be clearer.

felixguendling:
> > I think we need an option in `utl` to enable the docs. `UTL_BUILD_DOCS` with default value `OFF`.
> > And then wrap this in `if (UTL_BUILD_DOCS)`: [motis-project/utl@`69e700e`/CMakeLists.txt#L33](https://github.com/motis-project/utl/blob/69e700e03ae3d385304d754cf57877528bc591ef/CMakeLists.txt#L33)
> 
> Do all the MOTIS repos builds rely on the `CMakeLists.txt` in the `utl` module?
> 
> I'm not a big CMake expert, so I'm not sure how best to share the doc-building logic between MOTIS repositories. I think the following could be shared from `motis-project/utl`:
> 
> * `docs/CMakeLists.txt`
> * docs/Doxyfile (except from the module name)
> * `docs/conf.py` (except from the module name)
> * `docs/motis-logo.svg`
> * `docs/requirements.txt`
> * the `doc` job from `.github/workflows/unix.yml`
> 
> Maybe this should be tackled in a separate issue / PR?

In case this docs building is always the same and should be used in all MOTIS repositories (osr, motis, etc.), I would propose to create a separate repository and include it via `.pkg`. As long as it has a central `CMakeLists.txt` (then in the root directory) it will work. I think this is maybe a better solution than to force each repo to include `utl` if it wants only docs. Maybe at the moment all repos include `utl` but I would say it's maybe better to split it?

Lucas-C:
> I would propose to create a separate repository (maybe called `docs-gen` or simply `docs`) and include it via `.pkg`.
> I think this is maybe a better solution than to force each repo to include `utl` if it wants only docs.
> Maybe at the moment all repos include `utl` but I would say it's maybe better to split it?

I agree.
I will have to stop working on MOTIS for now, but I'll try to do this next week.
The documentation generation also requires Python and some depdencies to be installed with `pip`.
I'm not sure how best to do this from CMake, so unless you have some guidance regarding this, I'll search for existing projects doing something similar.

Lucas-C:
I see 3 options regarding how we deal with document generation in MOTIS:
1. we get it out of CMake managed tasks: it's maybe the simplest solution
2. we keep it inside CMake managed tasks, but make it optional
3. we keep it inside CMake managed tasks, and make it mandatory. This will require all MOTIS builds to have Doxygen & Python setup, so that's maybe a solution to exclude.

AMHA option 1) is maybe the best one for now, and would avoid adding some complexity to the `CMakeLists.txt` files.
What do you think?

We could still rely on `pkg` to have `motis-project/docs` installed in `deps`,
and then call `deps/docs/build_docs.sh` from a GitHub Actions `doc` job.

Also, I was wonder if you would be open to adopt [reusable workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows) in MOTIS, to avoid duplication of the YAML workflows (like the `doc` job)?

felixguendling:
The purpose of tracing in RAPTOR `include/nigiri/routing/raptor/debug.h` is mainly to be dumped to a text file that can become gigabytes in size for a single earliest arrival query (not even range query!). We neither want this stuff to go OTEL nor do we need timestamps, log levels, etc. here. I would like to keep it `fmt::println` for this reason. It's for MOTIS developers who need to go very very deep and check millions of routing operations to find the needle in the haystack if results don't match what we expect (e.g. from previous versions, or other implementations).

Lucas-C:
I am currently blocked by this issue: https://github.com/motis-project/utl/pull/35

felixguendling:
should be fixed by https://github.com/motis-project/utl/pull/36

> I am currently blocked by this issue: [motis-project/utl#35](https://github.com/motis-project/utl/pull/35)

Lucas-C:
I think this is ready

felixguendling:
I think you were right with the function signature. It's cluttering output a lot. Even if it gives additional context, it's maybe not worth to clutter the text output so much.

PR #173, opened 2025-02-10
maximum search interval

mority:
Adds an option to specify a maximum search interval per query, i.e., an interval that may not be exceeded by interval extensions during the search. Once the maximum interval has been searched the search will terminate even if the minimum connection count has not been reached.

### DIFF
--- a/include/nigiri/routing/interval_estimate.h
+++ b/include/nigiri/routing/interval_estimate.h
@@ -127,6 +127,10 @@ private:
     itv.to_ = tt_.external_interval().clamp(itv.to_);
     itv.from_ = data_type_max_interval_.clamp(itv.from_);
     itv.to_ = data_type_max_interval_.clamp(itv.to_);
+    if (q_.max_interval_) {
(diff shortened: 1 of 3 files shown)


felixguendling:
still needed in the ODM case?
I guess we'll have to set `extend_interval_earlier = extend_interval_later = false` anyway to have full control one layer above?

mority:
> still needed in the ODM case? I guess we'll have to set `extend_interval_earlier = extend_interval_later = false` anyway to have full control one layer above?

I thought about resolving the issue of the mismatching searched intervals of the sub-queries with a two-step process.
1. Run all sub-queries with min. connection count set and allow extension but set the max. interval to the interval where we have ODM rides prepared for, so that nigiri does not extend into a time region where no ODM connections can be found anymore.
2. Determine the largest interval searched by a sub-query. Rerun only the sub-queries that searched a smaller interval in the first step. For the re-runs we set the search interval to the largest interval from step 1 and do not allow extension.

PR #174, opened 2025-02-13
Improve trip display name

Altonss:
Related to https://github.com/public-transport/transitous/issues/886

### DIFF
--- a/src/loader/gtfs/trip.cc
+++ b/src/loader/gtfs/trip.cc
@@ -246,8 +246,8 @@ std::string trip::display_name(timetable const& tt) const {
                                std::string_view{route_->long_name_},
                                std::string_view{short_name_}};
 
-  // prefer route short name over route long name
-  // prefer route long name over trip short name,
+  // prefer route short name over route long name,
+  // prefer route long name over trip short name
   for (auto const candidate : precedence) {
     if (!candidate.empty()) {
       return std::string(candidate);
--- a/src/loader/gtfs/trip.cc
+++ b/src/loader/gtfs/trip.cc
@@ -242,28 +242,12 @@ std::string trip::display_name(timetable const& tt) const {
                        utl::parse<int>(short_name_));
   }
 
-  // take the first thing that starts with a letter and ends with a number
-  auto const starts_with_letter_and_ends_with_number =
-      [=](std::string_view line_id) {
-        return !is_digit(line_id.front()) && is_digit(line_id.back());
-      };
+  auto precedence = std::array{std::string_view{route_->short_name_},
+                               std::string_view{route_->long_name_},
+                               std::string_view{short_name_}};
 
-  auto precedence = std::array{std::string_view{short_name_},
-                               std::string_view{route_->short_name_},
-                               std::string_view{route_->long_name_}};
-
-  for (auto const candidate : precedence) {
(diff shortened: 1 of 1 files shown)


src/loader/gtfs/trip.cc
```diff
@@ -242,28 +242,12 @@ std::string trip::display_name(timetable const& tt) const {
                        utl::parse<int>(short_name_));
   }
 
-  // take the first thing that starts with a letter and ends with a number
-  auto const starts_with_letter_and_ends_with_number =
-      [=](std::string_view line_id) {
-        return !is_digit(line_id.front()) && is_digit(line_id.back());
-      };
+  auto precedence = std::array{std::string_view{route_->short_name_},
```
jbruechert, line 245: This doesn't work for all feeds that use semantic naming instead of the google maps convention.
That is as far as I know most of the data for Germany and Austria

felixguendling:
Bump. Has anyone tested this? It's still in draft mode.

src/loader/gtfs/trip.cc
jbruechert: as far as I remember for example the ÖBB feed entirely depends on this starts-with-letter-ends-with-number heuristic to display the right string

felixguendling: If there are only very few feeds with this problem we could also try to contact them and ask them if it's possible to write something useful into `route_name_short`. I would see Google as the inventor of GTFS as kind of a aid in interpreting the standard and point them to the Google way of doing things as "this is how it's done".

jbruechert:
I haven't tested it yet, but I think it doesn't catch all the cases. I think I know how to solve this properly (treat semantically-named feeds different from google maps feeds and have metadata on which is which), but I haven't gotten around to implementing it.

felixguendling:
If we consider the Google way to do this as kind of "the standard", we could just say "garbage in = garbage out" and try either to preprocess data outside of MOTIS or better: ask feed producers that produce non-compliant feeds to improve their data?

felixguendling:
In the end this PR would probably fix it in more places than it destroys something that worked, right? So overall it could still be an improvement to merge this PR as it is?

jbruechert:
Hmm yes maybe. I guess we should talk to Mobility Data to get the official interpretation. If they agree with the Google way I'm fine with doing it this way (and adding a gtfsclean pass to make the other feeds compliant in Transitous)

felixguendling:
If I remember correctly, there has been a Google Maps resource that outlined how they do it? I think this would be a good starting point for an issue and potential PR on https://github.com/google/transit

felixguendling:
I merge it now because I would say that it's already an improvement regarding the status quo in terms of "it fixes more than it destroys" (which doesn't mean that it destroys nothing but overall we improve the situation).

felixguendling:
I am open for both options: adding logic in nigiri or fixing the data before nigiri loads it.

jbruechert:
> But I can't find the link anymore..

This page was discussed previously:
https://developers.google.com/transit/gtfs/data-modeling/route-modeling-guide

jbruechert:
image

Some clicking through google maps later:
- they probably just display the route_short_name || route_long_name
- SNCF misuses the trip_headsign field which we are also seeing to display a trip name in google maps

PR #183, opened 2025-03-20
adding files for multi-criteria raptor

PumarPap:
added three files mcraptor.h and mcraptor.h/.cc

be aware: none of them are listet in CMakeLists yet


felixguendling:
- formatting
- missing tests

PR #186, opened 2025-04-09
RT ADDED/NEW/REPLACEMENT/DUPLICATED

traines-source:


include/nigiri/loader/gtfs/route.h
```diff
@@ -28,6 +28,8 @@ struct route {
 
 using route_map_t = hash_map<std::string, std::unique_ptr<route>>;
 
+clasz to_clasz(std::uint16_t);
```
felixguendling, line 31: in the global context, maybe `route_type_to_class` gives more information about the parameter than `std::uint16_t`

include/nigiri/rt/rt_timetable.h
```diff
@@ -35,11 +38,15 @@ using change_callback_t =
...
       timetable const&,
-      transport const,
-      std::span<stop::value_type> const& stop_seq = {},
-      std::span<delta_t> const& time_seq = {});
+      transport,
+      std::span<stop::value_type> const stop_seq = {},
+      std::span<delta_t> const time_seq = {},
+      std::string_view const new_trip_id = {},
+      std::string_view const route_id = {},
+      std::string_view const display_name = {},
```
felixguendling, line 48:
```suggestion
      std::span<stop::value_type> stop_seq = {},
      std::span<delta_t> time_seq = {},
      std::string_view new_trip_id = {},
      std::string_view route_id = {},
      std::string_view display_name = {},
```

`span` and `string_view` are const anyway (no non-const acessors)

include/nigiri/rt/rt_timetable.h
```diff
@@ -169,20 +176,25 @@ struct rt_timetable {
...
   vector_map<rt_transport_idx_t, variant<transport, rt_add_trip_id_idx_t>>
       rt_transport_static_transport_;
 
-  // RT trip ID index -> ID strings + source
-  vecvec<rt_add_trip_id_idx_t, char> trip_id_strings_;
+  // Lookup: additional external trip ID -> realtime transport
+  hash_map<string, rt_transport_idx_t> additional_trips_lookup_;
+
+  // RT trip ID index -> external trip ID
+  vecvec<rt_add_trip_id_idx_t, char> rt_add_trip_ids_;
```
felixguendling, line 187: maybe `string_store<rt_add_trip_id_idx_t>` would be a shortcut here?

include/nigiri/rt/gtfsrt_resolve_run.h
```diff
@@ -15,6 +17,7 @@ std::pair<run, trip_idx_t> gtfsrt_resolve_run(
     timetable const&,
     rt_timetable const*,
     source_idx_t,
-    transit_realtime::TripDescriptor const&);
+    transit_realtime::TripDescriptor const&,
+    std::string_view rt_trip_id = {});
```
felixguendling, line 21: Is this only for additional trips? then maybe call it rt_additional_trip_id (or similar)?

src/lookup/get_transport_stop_tz.cc
```diff
@@ -12,7 +12,9 @@ timezone_idx_t get_transport_stop_tz(timetable const& tt,
     return location_tz;
   }
 
-  auto const provider_idx = tt.transport_section_providers_[t].front();
+  auto const provider_idx = t != transport_idx_t::invalid()
+                                ? tt.transport_section_providers_[t].front()
+                                : provider_idx_t{0};  // TODO
```
felixguendling, line 17: what happens here?

src/rt/frun.cc
```diff
@@ -109,7 +111,13 @@ trip_idx_t run_stop::get_trip_idx(event_type const ev_type) const noexcept {
 
 std::string_view run_stop::trip_display_name(
     event_type const ev_type) const noexcept {
-  return tt().trip_display_names_[get_trip_idx(ev_type)].view();
+  if (fr_->is_scheduled()) {
+    return tt().trip_display_names_[get_trip_idx(ev_type)].view();
+  }
+  if (rtt() != nullptr) {
+    return rtt()->transport_name(tt(), fr_->rt_);
+  }
```
felixguendling, line 119: first checking rtt would allow GTFS-RT to override the trip name?

src/rt/gtfsrt_resolve_run.cc
```diff
@@ -96,10 +96,24 @@ void resolve_static(date::sys_days const today,
...
+void resolve_rt(rt_timetable const& rtt,
+                run& output,
+                std::string_view trip_id) {
   auto const it = rtt.static_trip_lookup_.find(output.t_);
   if (it != end(rtt.static_trip_lookup_)) {
     output.rt_ = it->second;
+    return;
+  }
+  auto const rtt_it =
+      rtt.additional_trips_lookup_.find(trip_id);  // TODO only check for added?
```
felixguendling, line 108: should only be necessary if `!trip_id.emtpy()`?

src/rt/gtfsrt_update.cc
```diff
@@ -149,41 +157,153 @@ unixtime_t fallback_pred(rt_timetable const& rtt,
...
+    }
+    return {};
+  };
+  auto const display_name =
+      tripUpdate.has_trip_properties() &&
+              tripUpdate.trip_properties().has_trip_short_name()
+          ? std::string_view{tripUpdate.trip_properties().trip_short_name()}
+          : std::string_view{};
+  // ADDED/NEW stops+times+new_trip_id, REPLACEMENT stops+times, DUPL
+  // new_trip_id
```
felixguendling, line 250: maybe split in 3 lines?

src/rt/gtfsrt_update.cc
```diff
@@ -149,41 +157,153 @@ unixtime_t fallback_pred(rt_timetable const& rtt,
...
+                  "absolute times are required for unscheduled trips");
+      utl::verify(stu.has_stop_id(),
+                  "stop_id is required for unscheduled trips");
+      auto const it =
+          tt.locations_.location_id_to_idx_.find({stu.stop_id(), src});
+      if (it == end(tt.locations_.location_id_to_idx_)) {
+        log(log_lvl::error, "rt.gtfs.unsupported",
+            "NEW/ADDED stop_id must be contained in stops.txt (src={}, "
+            "trip_id={}, "
+            "stop_id={}), skipping",
```
felixguendling, line 195: strange formatting

src/rt/rt_timetable.cc
```diff
@@ -1,27 +1,46 @@
...
     source_idx_t const src,
     timetable const& tt,
     transport const t,
-    std::span<stop::value_type> const& stop_seq,
-    std::span<delta_t> const& time_seq) {
+    std::span<stop::value_type> const stop_seq,
+    std::span<delta_t> const time_seq,
+    std::string_view const new_trip_id,
+    std::string_view const route_id,
+    std::string_view const display_name,
```
felixguendling, line 14:
```suggestion
    std::span<stop::value_type> stop_seq,
    std::span<delta_t> time_seq,
    std::string_view new_trip_id,
    std::string_view route_id,
    std::string_view display_name,
```

src/rt/gtfsrt_update.cc
```diff
@@ -149,41 +157,153 @@ unixtime_t fallback_pred(rt_timetable const& rtt,
...
+                 gtfsrt::TripUpdate const& tripUpdate) {
+  auto const& stus = tripUpdate.stop_time_update();
+  auto const sr = tripUpdate.trip().schedule_relationship();
+  auto const added_or_replaced =
+      is_added(sr) ||
+      sr == transit_realtime::TripDescriptor_ScheduleRelationship_REPLACEMENT;
+
+  auto stops = std::vector<stop::value_type>{};
+  if (added_or_replaced) {
+    for (auto& stu : stus) {
```
felixguendling, line 183:
```suggestion
    for (auto const& stu : stus) {
```

src/lookup/get_transport_stop_tz.cc
traines-source: Since this is only used for the TZ for debug printing so far, this was the easiest solution. But now that we store the provider with the route id anyways, I can make it more sophisticated

include/nigiri/rt/gtfsrt_resolve_run.h
traines-source: Yes this is only for DUPLICATED trips actually (i.e. not used rn)

src/rt/gtfsrt_resolve_run.cc
traines-source: It should never be empty, because except for DUPLICATED it will be the usual `trip_id`, but it's not straightforward to distinguish whether it could be a NEW/ADDED trip (e.g. when coming from the API), would need an additional flag when coming from gtfsrt_update I think. But I haven't really seen any performance hits, so...

src/loader/build_footpaths.cc
```diff
@@ -297,14 +297,15 @@ void connect_components(timetable& tt,
...
+        auto const adjusted_int = static_cast<int>(distance / kWalkSpeed / 60);
         if (adjusted_int > std::numeric_limits<u8_minutes::rep>::max()) {
           log(log_lvl::error, "loader.footpath.adjust",
               "too long after adjust: {}>256", adjusted_int);
+        } else {
+          adjusted = u8_minutes{
+              std::max(static_cast<duration_t::rep>(duration.count()),
+                       static_cast<duration_t::rep>(adjusted_int))};
         }
-        adjusted = u8_minutes{adjusted_int};
```
traines-source, line 307: Stumbled upon a footpath of 62996.7 minutes somewhere in Transitous (probably to 0,0?), i.e. even overflowed the `duration_t` before the check. Should we set to u8_max or keep the old value?

felixguendling: I guess overflow does not make sense, so clamping to u8 range makes sense?

PR #191, opened 2025-04-23
VDV RT: Run Cancelation & Stop Change

mority:


src/rt/rt_timetable.cc
```diff
@@ -76,4 +76,21 @@ rt_transport_idx_t rt_timetable::add_rt_transport(
   return rt_transport_idx_t{rt_t_idx};
 }
 
+void rt_timetable::cancel_run(rt::run& r) {
```
felixguendling, line 79: const?

felixguendling (approved):

src/rt/vdv/vdv_update.cc
```diff
@@ -24,7 +24,7 @@
 
 namespace nigiri::rt::vdv {
 
-// #define VDV_DEBUG
+#define VDV_DEBUG
```
felixguendling, line 27: default should be with or without logging?

PR #198, opened 2025-05-12
Display tag on import progress

traines-source:
depends on https://github.com/motis-project/utl/pull/39

many changes for a small improvement, alternative: display file name

### DIFF
--- a/exe/import.cc
+++ b/exe/import.cc
@@ -87,13 +87,14 @@ int main(int ac, char** av) {
       if (is_directory(e) /* unpacked zip file */ ||
           boost::algorithm::to_lower_copy(
               e.path().extension().generic_string()) == ".zip") {
-        input_files.emplace_back(e.path().generic_string(),
-                                 e.path().filename().generic_string(), c);
+        input_files.emplace_back(
+            timetable_source{e.path().filename().generic_string(),
+                             e.path().generic_string(), c});
       }
     }
   } else if (exists(in) && !recursive) {
-    input_files.emplace_back(in.generic_string(),
-                             in.filename().generic_string(), c);
+    input_files.emplace_back(timetable_source{in.filename().generic_string(),
+                                              in.generic_string(), c});
   }
 
   if (input_files.empty()) {
--- a/exe/import.cc
+++ b/exe/import.cc
@@ -81,8 +81,7 @@ int main(int ac, char** av) {
     return 0;
   }
 
-  auto input_files =
-      std::vector<std::tuple<std::string, std::string, loader_config>>{};
+  auto input_files = std::vector<timetable_source>{};
   if (is_directory(in) && recursive) {
     for (auto const& e : fs::directory_iterator(in)) {
       if (is_directory(e) /* unpacked zip file */ ||
--- a/exe/import.cc
+++ b/exe/import.cc
@@ -81,17 +81,20 @@ int main(int ac, char** av) {
     return 0;
   }
 
-  auto input_files = std::vector<std::pair<std::string, loader_config>>{};
+  auto input_files =
+      std::vector<std::tuple<std::string, std::string, loader_config>>{};
   if (is_directory(in) && recursive) {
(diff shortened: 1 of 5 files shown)


felixguendling:
Why can't we set the tag 1x on the progress bar and it will display this tag until a new tag is set?

traines-source:
You're right, I was able to move this up one layer of abstraction.

src/loader/load.cc
```diff
@@ -49,9 +52,13 @@ timetable load(std::vector<std::pair<std::string, loader_config>> const& paths,
       if (!is_in_memory) {
         log(log_lvl::info, "loader.load", "loading {}", path);
       }
+      auto const progress_tracker = utl::get_active_progress_tracker();
+      progress_tracker->context(std::string{tag});
       try {
         (*it)->load(local_config, src, *dir, tt, bitfields, a, shapes);
+        progress_tracker->context("");
       } catch (std::exception const& e) {
+        progress_tracker->context("");
```
felixguendling, line 61: would this lead to *not*  displaying the context in case something fails (when I would probably be most interested in knowing what the tag is)?

src/loader/load.cc
```diff
@@ -22,12 +23,14 @@ std::vector<std::unique_ptr<loader_interface>> get_loaders() {
...
 }
 
-timetable load(std::vector<std::pair<std::string, loader_config>> const& paths,
-               finalize_options const& finalize_opt,
-               interval<date::sys_days> const& date_range,
-               assistance_times* a,
-               shapes_storage* shapes,
-               bool ignore) {
+timetable load(
+    std::vector<std::tuple<std::string, std::string, loader_config>> const&
```
felixguendling, line 27:
```suggestion
    std::vector<std::tuple<std::string, std::string, loader_config>> const&
```

maybe a struct would make it more obvious and help to not swap the `std::string` parameters in the tuple by accident?

src/loader/load.cc
```diff
@@ -49,9 +52,13 @@ timetable load(std::vector<std::pair<std::string, loader_config>> const& paths,
       if (!is_in_memory) {
         log(log_lvl::info, "loader.load", "loading {}", path);
       }
+      auto const progress_tracker = utl::get_active_progress_tracker();
+      progress_tracker->context(std::string{tag});
       try {
         (*it)->load(local_config, src, *dir, tt, bitfields, a, shapes);
+        progress_tracker->context("");
```
felixguendling, line 59: we would also have utl RAII stuff for this (to make sure that once the scope exits this code is executed, no matter what)

src/loader/load.cc
traines-source: The path should be shown at least, but yes you're right it doesn't really make sense to reset it, assuming everything dies.

felixguendling (approved):

exe/import.cc
```diff
@@ -81,17 +81,20 @@ int main(int ac, char** av) {
...
 
-  auto input_files = std::vector<std::pair<std::string, loader_config>>{};
+  auto input_files = std::vector<timetable_source>{};
   if (is_directory(in) && recursive) {
     for (auto const& e : fs::directory_iterator(in)) {
       if (is_directory(e) /* unpacked zip file */ ||
           boost::algorithm::to_lower_copy(
               e.path().extension().generic_string()) == ".zip") {
-        input_files.emplace_back(e.path().generic_string(), c);
+        input_files.emplace_back(
```
felixguendling, line 90: if emplace_back doesn't really have any advantage over push_back like it's the case here (I think), it would make more sense to use push_back (less template instantiations = faster compilation + smaller binary size)

exe/import.cc
```diff
@@ -81,17 +81,20 @@ int main(int ac, char** av) {
...
+            timetable_source{e.path().filename().generic_string(),
```
felixguendling, line 91: `.field_ = ` would make it more obvious that fields are in the right order without checking the struct definition (especially since here the types would also match in the wrong order)

PR #199, opened 2025-05-13
Extend direction guessing for scheduled trips

jbruechert:


src/rt/frun.cc
```diff
@@ -167,14 +167,15 @@ std::string_view run_stop::direction(event_type const ev_type) const noexcept {
...
-        .name();
+
+    if (!direction.empty()) {
+      return direction;
+    }
   }
-  return "";
+
+  return run_stop{.fr_ = fr_,
+                  .stop_idx_ = static_cast<stop_idx_t>(fr_->size() - 1U)}
```
traines-source, line 177: This crashes for ADDED trips when the `rtt` timetable hasn't been passed to the `frun`, which happened somewhere I think (although I have probably fixed that). That's why I added the `rtt() != nullptr` check to be sure, now probably `fr_->is_scheduled() || rtt() != nullptr`.

jbruechert: Thanks for the hint, should be fixed now

felixguendling (approved):

traines-source:
Thanks! I'll update the nigiri dep over in https://github.com/motis-project/motis/pull/858.

PR #204, opened 2025-05-29
parse gtfs stop_desc

hannsadrian:
this adds the import for stop_desc in the gtfs stops.txt as description of locations.


felixguendling:
Maybe add some checks to https://github.com/motis-project/nigiri/blob/master/test/loader/gtfs/stop_test.cc so we avoid regressions if someone refactors something.

PR #207, opened 2025-05-29
Add support for importing platform_code of stations/stops without a parent

hannsadrian:
This pr adds the ability to import GTFS stops that have a platform_code, but no parent_station. It does this by creating artificial parents which hold the station's name, as motis locations only have name and id but no field for track information. 

stops.txt entries with platform_code but no parent_station is not inline with the GTFS spec, but common practice at least for the Delfi feed. With this pr we strengthen the import to support track data, even if it's a bit out of spec.

### DIFF
--- a/src/loader/gtfs/stop.cc
+++ b/src/loader/gtfs/stop.cc
@@ -229,10 +229,10 @@ stops_map_t read_stops(source_idx_t const src,
                                    []() { return std::make_unique<stop>(); })
                     .get();
             parent->id_ = new_stop->id_;
+            parent->name_ = new_stop->name_;
             parent->coord_ = new_stop->coord_;
             parent->desc_ = new_stop->desc_;
             parent->timezone_ = new_stop->timezone_;
-            parent->platform_code_ = "";
             parent->children_.emplace(new_stop);
             new_stop->parent_ = parent;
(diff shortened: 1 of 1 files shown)


src/loader/gtfs/stop.cc
```diff
@@ -245,8 +265,11 @@ stops_map_t read_stops(source_idx_t const src,
   }
 
   auto empty_idx_vec = vector<location_idx_t>{};
-  for (auto const& [id, s] : stops) {
+  for (auto const& [id_pair, s] : stops) {
     auto const is_track = s->parent_ != nullptr && !s->platform_code_.empty();
+    // if the second entry of the pair is true, we have an artificial parent
+    auto const id =
+        std::string{id_pair.first} + (id_pair.second ? "-motisparent" : "");
```
felixguendling, line 272: use `fmt::format`

hannsadrian:
Thanks for your feedback, should now be improved

felixguendling (approved):

traines-source (approved):
Sorry for the wait, I overlooked this. Looking really good, indeed quite a few new platform labels have appeared for Germany (of course also some wrong/misleading ones...)!

PR #208, opened 2025-05-29
Make route_short_name and trip_short_name available individually

jbruechert:


src/timetable.cc
```diff
@@ -23,6 +23,21 @@ void timetable::resolve() {
   }
 }
 
+std::string_view timetable::route_short_name(const transport_idx_t t) const {
+  auto const trip_idx =
+      merged_trips_[transport_to_trip_section_[t].front()].front();
```
felixguendling, line 28: one transport can be N concatenated trips via `block_id`
so for debugging purposes, it's sometimes okay to only return the first
but for the API we want the correct route/trip name to be displayed (on the map, etc)

jbruechert:
Still pretty unsure about the rt_timetable, need to test with some real data later.

src/rt/frun.cc
```diff
@@ -99,14 +99,25 @@ trip_idx_t run_stop::get_trip_idx(event_type const ev_type) const {
...
   if (fr_->is_scheduled()) {
-    return tt().trip_display_names_[get_trip_idx(ev_type)].view();
+    return tt().route_short_name(get_trip_idx(ev_type));
   }
-  return "?";
+  return {};
+}
+
+std::string_view run_stop::trip_short_name(
+    event_type const event_type) const noexcept {
```
felixguendling, line 113:
```suggestion
    event_type const event_type) const {
```

I guess `noexcept` is correct here but since we had several crashes where I thought no exception can be raised and MOTIS terminated completely, I decided to not use `noexcept` except it's in hot loops and improves performance.

src/timetable.cc
```diff
@@ -23,6 +23,17 @@ void timetable::resolve() {
...
 }
 
+std::string_view timetable::route_short_name(const trip_idx_t trip_idx) const {
+  const auto source_idx = trip_id_src_[trip_ids_[trip_idx].front()];
+  return route_ids_[source_idx]
+      .route_id_short_names_[trip_route_id_[trip_idx]]
+      .view();
+}
+
+std::string_view timetable::trip_short_name(const trip_idx_t trip_idx) const {
```
felixguendling, line 33:
```suggestion
std::string_view timetable::trip_short_name(trip_idx_t const trip_idx) const {
```

src/timetable.cc
```diff
@@ -23,6 +23,17 @@ void timetable::resolve() {
   }
 }
 
+std::string_view timetable::route_short_name(const trip_idx_t trip_idx) const {
```
felixguendling, line 26:
```suggestion
std::string_view timetable::route_short_name(trip_idx_t const trip_idx) const {
```

src/rt/frun.cc
felixguendling, line 113:
```suggestion
    event_type const ev_type) const noexcept {
```

jbruechert: Ah yes, I just messed up the rebase here.

PR #227, opened 2025-07-10
VDV AUS: Tweaks

mority:
- stricter threshold for additional matches to prevent false positives in case of low headway: 0.9 --> 0.975
- improved event time matching: 1st stop: match dep, last stop: match arr, else: match any


mority:
Hm, still seeing false positives for multiple matches. I will generate a test case from the feed data and see how far the additional match threshold has to be increased to prevent this.

PR #234, opened 2025-07-25
Add timetable imort statistics

MichaelKutzner:
Computes and stores timetable statistics, as suggested in motis-project/motis#935

I still don't like the way the JSON output is generated. Maybe a raw format would be  better for this? I'm also not sure, if adding the source path will be helpful for the metrics endpoint. Or if I should remove it.


felixguendling (commented):
I would've had expected the following three functions:

In nigiri:
- `get_metrics(nigiri::timetable) -> nigiri::timetable_metrics`
- `to_str(nigiri::timetable_metrics const&)`

In MOTIS:
- something to convert `nigiri::timetable_metrics` to prometheus

In MOTIS, statistics would be written to the `data` folder in this block:

```cpp
        if (write) {
          d.tt_->write(data_path / "tt.bin");
          d.tags_->write(data_path / "tags.bin");
          std::ofstream{data_path / "tt.metrics.json"} << n::get_metrics(*d.tt_);
        }
```

There's no real need to do this intrusive and mix metrics and `load_timetable` code as this would be specific to GTFS and would need to be rewritten for NeTEx + HRDF. All information to generate the metrics should be in the timetable. If not, we can add some more fields.

There's also no need to store the metrics in the timetable itself as the information can be derived.

MichaelKutzner:
The code has been updates to compute the metrics without the need to store any additional fields. The keys are now easier to type as well.

There are still some points, where I'm not sure, if this is the best approach:
* Starting with `trip_idx` and `trip_id`, merged trips might be counted multiple times. Would be an issue?
* `to_str()` requires a `timetable` reference to get the correct days. To remove this, the metrics could store `date::sys_days` instead, but this requires more memory.
* Some names are a bit verbose.

include/nigiri/timetable_metrics.h
```diff
@@ -0,0 +1,24 @@
...
+
+namespace nigiri {
+
+struct timetable_metrics {
+  struct feed_metrics {
+    std::uint64_t transport_days_;
+    std::uint32_t locations_;
+    std::uint32_t trips_;
+    std::uint16_t first_ = std::numeric_limits<std::uint16_t>::max();
+    std::uint16_t last_ = std::numeric_limits<std::uint16_t>::min();
```
felixguendling, line 16: for consistency always use uniform aggregate initialization / uniform initialization as described in the style guide

https://en.cppreference.com/w/cpp/language/aggregate_initialization.html

felixguendling (approved):

include/nigiri/timetable_metrics.h
MichaelKutzner: Thank you for the reminder. I have updated this.

src/timetable_metrics.cc
```diff
@@ -0,0 +1,91 @@
...
+#include "fmt/format.h"
+
+#include "date/date.h"
+
+namespace nigiri {
+
+struct transport_day_metric {
+  std::uint16_t first_ = std::numeric_limits<std::uint16_t>::max();
+  std::uint16_t last_ = std::numeric_limits<std::uint16_t>::min();
+  std::size_t count_ = 0U;
```
felixguendling, line 16: same

src/timetable_metrics.cc
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  return m;
+}
+
+std::string to_str(timetable_metrics const& m, timetable const& tt) {
+  auto const from = std::chrono::time_point_cast<date::sys_days::duration>(
+      tt.internal_interval().from_);
+  std::stringstream ss{};
```
felixguendling, line 76: AAA style (see style guide)

MichaelKutzner: Sorry, missed that. It's fixed now.

PR #238, opened 2025-07-28
reduce footpaths

felixguendling:


src/loader/reduce_footpaths.cc
```diff
@@ -0,0 +1,107 @@
...
+        auto const l = location_idx_t{l_idx};
+
+        if (s.route_fps_.size() != tt.n_locations()) {
+          s.route_fps_ = vector_map<route_idx_t, std::array<footpath, kN>>{
+              tt.n_locations(), init};
+          s.route_is_reachable_.resize(tt.n_locations());
+        }
+
+        // Group by route (duplicates footpaths).
+        // This eliminates footpaths to locations w/o scheduled traffic.
```
miklcct, line 43: How about footpaths to locations which doesn't have scheduled traffic, but are intended to be used for real-time trips, for example, stops reserved by rail replacement buses or spare rail platforms?

felixguendling: They will be recovered from `footpaths_full_out_` which stores the full set of footpaths (unfiltered) and have to be copied to the `rt_timetable` in case a previously unused stop will be visited.

The main goal of this branch was to figure out the speedup potential. Results so far:

- negligible speedup for default footpaths
- measurable speedup for 20min walking footpaths
- to be measured for longer settings (probably speedup will be higher the greater the distance, but still calculation time will grow in total)

PR #241, opened 2025-07-29
prototype for the compution of an estimated delay of vehicles

abantobon6:
This code uses GTFS-RT feed messages of type ‘VehiclePosition’ and utilises the GPS data within to determine an estimated delay for vehicles.

### DIFF
--- a/include/nigiri/rt/gtfsrt_update.h
+++ b/include/nigiri/rt/gtfsrt_update.h
@@ -52,19 +52,22 @@ statistics gtfsrt_update_msg(timetable const&,
                              rt_timetable&,
                              source_idx_t const,
                              std::string_view tag,
-                             transit_realtime::FeedMessage const&);
+                             transit_realtime::FeedMessage const&,
+                             bool use_vp);
 
 statistics gtfsrt_update_buf(timetable const& tt,
                              rt_timetable& rtt,
                              source_idx_t const src,
                              std::string_view tag,
                              std::string_view protobuf,
-                             transit_realtime::FeedMessage& msg);
+                             transit_realtime::FeedMessage& msg,
+                             bool use_vp);
 
 statistics gtfsrt_update_buf(timetable const&,
                              rt_timetable&,
                              source_idx_t const,
                              std::string_view tag,
-                             std::string_view protobuf);
+                             std::string_view protobuf,
+                             bool use_vp);
 
 }  // namespace nigiri::rt
--- a/include/nigiri/rt/gtfsrt_update.h
+++ b/include/nigiri/rt/gtfsrt_update.h
@@ -25,6 +25,7 @@ struct statistics {
   int total_entities_success_{0};
   int total_entities_fail_{0};
   int total_alerts_{0};
+  int total_vehicles_{0};
   int alert_total_informed_entities_{0};
   int alert_total_resolve_success_{0};
   int alert_trip_not_found_{0};
@@ -38,6 +39,7 @@ struct statistics {
   int unsupported_no_trip_id_{0};
   int no_vehicle_position_{0};
   int vehicle_position_without_position_{0};
+  int vehicle_position_without_timestamp_{0};
   int vehicle_position_without_trip_{0};
   int vehicle_position_trip_without_trip_id_{0};
   int vehicle_position_trip_without_route_id_{0};
@@ -53,7 +55,7 @@ statistics gtfsrt_update_msg(timetable const&,
                              source_idx_t const,
                              std::string_view tag,
                              transit_realtime::FeedMessage const&,
-                             bool use_vp);
+                             bool use_vp = false);
 
 statistics gtfsrt_update_buf(timetable const& tt,
                              rt_timetable& rtt,
@@ -61,13 +63,13 @@ statistics gtfsrt_update_buf(timetable const& tt,
                              std::string_view tag,
                              std::string_view protobuf,
                              transit_realtime::FeedMessage& msg,
-                             bool use_vp);
+                             bool use_vp = false);
 
 statistics gtfsrt_update_buf(timetable const&,
                              rt_timetable&,
                              source_idx_t const,
                              std::string_view tag,
                              std::string_view protobuf,
-                             bool use_vp);
+                             bool use_vp = false);
 
 }  // namespace nigiri::rt
--- a/include/nigiri/rt/gtfsrt_update.h
+++ b/include/nigiri/rt/gtfsrt_update.h
@@ -35,11 +35,9 @@ struct statistics {
   int alert_agency_id_not_found_{0};
   int alert_invalid_route_type_{0};
   int unsupported_deleted_{0};
-  int unsupported_vehicle_{0};
   int unsupported_no_trip_id_{0};
   int no_vehicle_position_{0};
   int vehicle_position_without_position_{0};
-  int vehicle_position_position_without_latlon_{0};
   int vehicle_position_without_trip_{0};
   int vehicle_position_trip_without_trip_id_{0};
   int vehicle_position_trip_without_route_id_{0};
--- a/include/nigiri/rt/gtfsrt_update.h
+++ b/include/nigiri/rt/gtfsrt_update.h
@@ -42,7 +42,10 @@ struct statistics {
   int vehicle_position_without_timestamp_{0};
   int vehicle_position_without_trip_{0};
   int vehicle_position_trip_without_trip_id_{0};
-  int vehicle_position_trip_without_route_id_{0};
+  int duplicate_vehicle_position_timestamps_{0};
+  int outdated_or_illegal_vehicle_position_timestamps_{0};
+  int illegal_stop_{0};
+  int vehicle_position_position_not_at_stop_{0};
   int no_trip_update_{0};
   int trip_update_without_trip_{0};
   int trip_resolve_error_{0};
--- a/include/nigiri/rt/gtfsrt_update.h
+++ b/include/nigiri/rt/gtfsrt_update.h
@@ -37,6 +37,12 @@ struct statistics {
   int unsupported_deleted_{0};
   int unsupported_vehicle_{0};
   int unsupported_no_trip_id_{0};
+  int no_vehicle_position_{0};
+  int vehicle_position_without_position_{0};
+  int vehicle_position_position_without_latlon_{0};
+  int vehicle_position_without_trip_{0};
+  int vehicle_position_trip_without_trip_id_{0};
+  int vehicle_position_trip_without_route_id_{0};
   int no_trip_update_{0};
   int trip_update_without_trip_{0};
   int trip_resolve_error_{0};
--- a/include/nigiri/rt/rt_timetable.h
+++ b/include/nigiri/rt/rt_timetable.h
@@ -204,6 +204,12 @@ struct rt_timetable {
   vecvec<rt_transport_idx_t, delta_t> rt_transport_stop_times_;
   vecvec<rt_transport_idx_t, stop::value_type> rt_transport_location_seq_;
 
+  // RT Transport -> latest known VehiclePosition timestamp
+  hash_map<rt_transport_idx_t, unixtime_t> rt_transport_latest_vehicle_time_;
+
+  // RT Transport -> latest known stop through VehiclePosition
+  hash_map<rt_transport_idx_t, stop_idx_t> rt_transport_latest_stop_;
+
   // RT trip index -> display name (empty if not changed)
   vecvec<rt_transport_idx_t, char> rt_transport_trip_short_names_;
   vecvec<rt_transport_idx_t, char> rt_transport_line_;
--- a/include/nigiri/types.h
+++ b/include/nigiri/types.h
@@ -298,6 +298,7 @@ using u8_minutes = std::chrono::duration<std::uint8_t, std::ratio<60>>;
 using duration_t = i16_minutes;
 using unixtime_t = std::chrono::sys_time<i32_minutes>;
 using local_time = date::local_time<i32_minutes>;
+using time_point_t = std::chrono::time_point<date::sys_days::clock, date::sys_days::duration>;
 
 constexpr u8_minutes operator""_u8_minutes(unsigned long long n) {
   return duration_t{n};
--- a/src/rt/gtfsrt_update.cc
+++ b/src/rt/gtfsrt_update.cc
@@ -18,9 +18,6 @@
 #include "nigiri/rt/gtfsrt_resolve_run.h"
 #include "nigiri/rt/run.h"
 
-// Anton war hier
-
-
 namespace gtfsrt = transit_realtime;
 namespace protob = google::protobuf;
 
@@ -80,8 +77,15 @@ std::ostream& operator<<(std::ostream& out, statistics const& s) {
                     s.alert_total_informed_entities_);
   print_if_no_empty("alert_invalid_route_type", s.alert_invalid_route_type_,
                     s.alert_total_informed_entities_);
-  print_if_no_empty("unsupported_vehicle", s.unsupported_vehicle_, true);
   print_if_no_empty("unsupported_no_trip_id", s.unsupported_no_trip_id_, true);
+  print_if_no_empty("no_trip_update", s.no_vehicle_position_, true);
+  print_if_no_empty("no_trip_update", s.vehicle_position_without_position_,
+                    true);
+  print_if_no_empty("no_trip_update", s.vehicle_position_without_trip_, true);
+  print_if_no_empty("no_trip_update", s.vehicle_position_trip_without_trip_id_,
+                    true);
+  print_if_no_empty("no_trip_update", s.vehicle_position_trip_without_route_id_,
+                    true);
   print_if_no_empty("no_trip_update", s.no_trip_update_, true);
   print_if_no_empty("trip_update_without_trip", s.trip_update_without_trip_,
                     true);
@@ -467,40 +471,30 @@ statistics gtfsrt_update_msg(timetable const& tt,
             R"(unsupported: no "position" field in "vehicle_position" field (tag={}, id={}), skipping message)",
             tag, entity.id());
         ++stats.vehicle_position_without_position_;
-      }
-      else if (!entity.vehicle().position().has_longitude() || !entity.vehicle().position().has_latitude()) {
-        log(log_lvl::error, "rt.gtfs.unsupported",
-            R"(unsupported: no "latitude" or "longitude" field in "location" field (tag={}, id={}), skipping message)",
-            tag, entity.id());
-        ++stats.vehicle_position_position_without_latlon_;
-      }
-      else if (!entity.vehicle().has_trip()) {
+      } else if (!entity.vehicle().has_trip()) {
         log(log_lvl::error, "rt.gtfs.unsupported",
             R"(unsupported: no "trip" field in "vehicle_position" field (tag={}, id={}), skipping message)",
             tag, entity.id());
         ++stats.vehicle_position_without_trip_;
-      }
-      else if (!entity.vehicle().trip().has_trip_id()) {
+      } else if (!entity.vehicle().trip().has_trip_id()) {
         log(log_lvl::error, "rt.gtfs.unsupported",
             R"(unsupported: no "trip_id" field in "trip" field (tag={}, id={}), skipping message)",
             tag, entity.id());
         ++stats.vehicle_position_trip_without_trip_id_;
-      }
-      else if (!entity.vehicle().trip().has_route_id()) {
+      } else if (!entity.vehicle().trip().has_route_id()) {
         log(log_lvl::error, "rt.gtfs.unsupported",
             R"(unsupported: no "route_id" field in "trip" field (tag={}, id={}), skipping message)",
             tag, entity.id());
         ++stats.vehicle_position_trip_without_route_id_;
-      }
-      else {
+      } else {
         auto const vp = entity.vehicle();
         auto const vp_lat = vp.position().latitude();
         auto const vp_lon = vp.position().longitude();
         auto const td = vp.trip();
         auto const gtfsrt_trip_id = td.trip_id();
 
-        auto [r, trip_idx] = gtfsrt_resolve_run(today, tt, &rtt, src, td,
-                                                std::string_view{gtfsrt_trip_id});
+        auto [r, trip_idx] = gtfsrt_resolve_run(
+            today, tt, &rtt, src, td, std::string_view{gtfsrt_trip_id});
         auto const& rtt_const = rtt;
         auto const location_seq =
             r.is_scheduled()
@@ -552,20 +546,21 @@ statistics gtfsrt_update_msg(timetable const& tt,
         // get remaining stops
         auto const seq_numbers =
             r.is_scheduled()
-                ? loader::gtfs::
-                      stop_seq_number_range{{tt.trip_stop_seq_numbers_[trip_idx]},
-                                            static_cast<stop_idx_t>(
-                                                r.stop_range_.size())}
+                ? loader::gtfs::stop_seq_number_range{{tt.trip_stop_seq_numbers_
+                                                           [trip_idx]},
+                                                      static_cast<stop_idx_t>(
+                                                          r.stop_range_.size())}
                 : loader::gtfs::stop_seq_number_range{
                       std::span<stop_idx_t>{},
                       static_cast<stop_idx_t>(location_seq.size())};
 
         // get delay
-        auto const now = vp.has_timestamp()
-                             ? unixtime_t{std::chrono::duration_cast<i32_minutes>(
-                                   std::chrono::seconds{vp.timestamp()})}
-                             : std::chrono::floor<i32_minutes>(
-                                   std::chrono::system_clock::now());
+        auto const now =
+            vp.has_timestamp()
+                ? unixtime_t{std::chrono::duration_cast<i32_minutes>(
+                      std::chrono::seconds{vp.timestamp()})}
+                : std::chrono::floor<i32_minutes>(
+                      std::chrono::system_clock::now());
         auto const delay =
             now - tt.event_time(r.t_, stopped_at_idx, event_type::kDep);
         auto const delay_cast = std::chrono::duration_cast<duration_t>(delay);
@@ -586,6 +581,7 @@ statistics gtfsrt_update_msg(timetable const& tt,
           }
         }
       }
+      continue;
     }
     log(log_lvl::error, "rt.gtfs.unsupported",
         R"(unsupported: no "vehicle_position" field (tag={}, id={}), skipping message)",
--- a/src/rt/gtfsrt_update.cc
+++ b/src/rt/gtfsrt_update.cc
@@ -525,7 +525,8 @@ statistics gtfsrt_update_msg(timetable const& tt,
                              rt_timetable& rtt,
                              source_idx_t const src,
                              std::string_view tag,
-                             gtfsrt::FeedMessage const& msg) {
+                             gtfsrt::FeedMessage const& msg,
+                             bool use_vp) {
   auto span = get_otel_tracer()->StartSpan("gtfsrt_update_msg", {{"tag", tag}});
   auto scope = opentelemetry::trace::Scope{span};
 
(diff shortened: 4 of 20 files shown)


src/rt/gtfsrt_update.cc
```diff
@@ -447,10 +455,132 @@ statistics gtfsrt_update_msg(timetable const& tt,
...
+          r.rt_ = rtt.add_rt_transport(
+              src, tt, r.t_, stops, times, std::string_view{gtfsrt_trip_id},
+              td.route_id(), tt.trip_display_names_[trip_idx].view());
+        }
+
+        // match position to stop
+        auto at_stop = false;
+        auto stopped_at_idx = 0;
+
+        for (auto const loc : location_seq) {
```
felixguendling, line 522: that's a `utl::find_if`

src/rt/gtfsrt_update.cc
```diff
@@ -447,10 +455,132 @@ statistics gtfsrt_update_msg(timetable const& tt,
...
+            R"(unsupported: no "trip_id" field in "trip" field (tag={}, id={}), skipping message)",
+            tag, entity.id());
+        ++stats.vehicle_position_trip_without_trip_id_;
+      } else if (!entity.vehicle().trip().has_route_id()) {
+        log(log_lvl::error, "rt.gtfs.unsupported",
+            R"(unsupported: no "route_id" field in "trip" field (tag={}, id={}), skipping message)",
+            tag, entity.id());
+        ++stats.vehicle_position_trip_without_route_id_;
+      } else {
+        auto const vp = entity.vehicle();
```
felixguendling, line 483: separate function (like `update_run`)

should also only be done if a flag from outside says to use `VehiclePosition` updates as most feeds that provide them also have `TripUpdates` which makes `VehiclePosition` delay predictions obsolete

src/rt/gtfsrt_update.cc
```diff
@@ -447,10 +455,132 @@ statistics gtfsrt_update_msg(timetable const& tt,
...
+              src, tt, r.t_, stops, times, std::string_view{gtfsrt_trip_id},
+              td.route_id(), tt.trip_display_names_[trip_idx].view());
+        }
+
+        // match position to stop
+        auto at_stop = false;
+        auto stopped_at_idx = 0;
+
+        for (auto const loc : location_seq) {
+          stop const stop_loc{loc};
```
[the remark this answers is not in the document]
felixguendling: see STYLE.md

src/rt/gtfsrt_update.cc
```diff
@@ -447,10 +455,132 @@ statistics gtfsrt_update_msg(timetable const& tt,
...
+                       std::nullopt);
+          if (++seq_it != seq_numbers.end()) {
+            update_delay(tt, rtt, r, stop_idx, event_type::kDep, delay_cast,
+                         std::nullopt);
+          }
+        }
+      }
+      continue;
+    }
+    log(log_lvl::error, "rt.gtfs.unsupported",
```
felixguendling, line 579: ?

test/main.cc
```diff
@@ -14,6 +14,13 @@
 namespace fs = std::filesystem;
 
 int main(int argc, char** argv) {
+
```
felixguendling, line 17: ???

src/rt/gtfsrt_update.cc
```diff
@@ -75,8 +76,15 @@ std::ostream& operator<<(std::ostream& out, statistics const& s) {
                     s.alert_total_informed_entities_);
   print_if_no_empty("alert_invalid_route_type", s.alert_invalid_route_type_,
                     s.alert_total_informed_entities_);
-  print_if_no_empty("unsupported_vehicle", s.unsupported_vehicle_, true);
   print_if_no_empty("unsupported_no_trip_id", s.unsupported_no_trip_id_, true);
+  print_if_no_empty("no_trip_update", s.no_vehicle_position_, true);
```
felixguendling, line 80: ???

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +427,106 @@ bool update_run(source_idx_t const src,
...
+  auto const& rtt_const = rtt;
+  auto const location_seq =
+      r.is_scheduled()
+          ? std::span{tt.route_location_seq_[tt.transport_route_[r.t_.t_idx_]]}
+          : std::span{rtt_const.rt_transport_location_seq_[r.rt_]};
+  auto stops = std::vector<stop::value_type>{};
+
+  for (auto const loc : location_seq) {
+    stops.emplace_back(stop{loc}.value());
+  }
```
felixguendling, line 453: avoid allocations and unnecessary copies at all cost
this is both

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +427,106 @@ bool update_run(source_idx_t const src,
...
+  // match position to stop
+  auto it = utl::find_if(location_seq, [&](auto const& loc) {
+    auto const stop_loc = stop{loc};
+    auto const stop_position = tt.locations_.get(stop_loc.location_idx());
+    auto const vehicle_position = geo::latlng{vp_lat, vp_lon};
+    return geo::approx_squared_distance(
+               stop_position.pos_, vehicle_position,
+               geo::approx_distance_lng_degrees(vehicle_position)) < 10;
+  });
+  auto stopped_at_idx = std::distance(begin(location_seq), it);
```
felixguendling, line 478: why not const?

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +427,106 @@ bool update_run(source_idx_t const src,
...
+    auto times = added_or_replaced
+                     ? std::vector<delta_t>(location_seq.size() * 2U - 2U, 0)
+                     : std::vector<delta_t>{};
+    r.rt_ = rtt.add_rt_transport(
+        src, tt, r.t_, stops, times, std::string_view{gtfsrt_trip_id},
+        td.route_id(), tt.trip_display_names_[trip_idx].view());
+  }
+
+  // match position to stop
+  auto it = utl::find_if(location_seq, [&](auto const& loc) {
```
felixguendling, line 470: why not const?

src/rt/gtfsrt_update.cc
felixguendling, line 478: distance can be skipped if it==end

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +427,106 @@ bool update_run(source_idx_t const src,
...
+      now - tt.event_time(r.t_, stopped_at_idx, event_type::kDep);
+  auto const delay_cast = std::chrono::duration_cast<duration_t>(delay);
+  // update delay for following stops
+  auto seq_it = begin(seq_numbers);
+  for (int i = 0; i <= stopped_at_idx; ++i) {
+    ++seq_it;
+  }
+
+  // update delay for remaining stops
+  for (auto stop_idx = stopped_at_idx + 1; seq_it != seq_numbers.end();
```
felixguendling, line 513: this doesn't handle transfer_type=4 and block_id transfers correctly?
also: are there any ways this can become a invalid time sequence (has to be monotonously increasing, otherwise the routing won't work properly)
multiple vehicle position messages can be in the GTFS-RT protobuf and this has to be handled gracefully even if it's not a valid/useful update

test/rt/util.cc
```diff
@@ -71,7 +71,7 @@ void with_rt_trips(
       }
     }
     rt::gtfsrt_update_msg(tt, rtt, source_idx_t{0}, "",
-                          to_feed_msg(trip_delays, base_day + 1h));
+                          to_feed_msg(trip_delays, base_day + 1h), false);
```
felixguendling, line 74:
```suggestion
                          to_feed_msg(trip_delays, base_day + 1h));
```

give the parameter a default value and you can leave out all these changes

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +427,106 @@ bool update_run(source_idx_t const src,
   return true;
 }
 
+void handle_vehicle_position(timetable const& tt,
+                             rt_timetable& rtt,
+                             source_idx_t const src,
+                             std::string_view tag,
+                             gtfsrt::FeedEntity const& entity,
+                             std::chrono::time_point<date::sys_days::clock, date::sys_days::duration> today) {
```
felixguendling, line 435: use shorthand

src/rt/gtfsrt_update.cc
```diff
@@ -76,8 +77,15 @@ std::ostream& operator<<(std::ostream& out, statistics const& s) {
...
-  print_if_no_empty("unsupported_vehicle", s.unsupported_vehicle_, true);
   print_if_no_empty("unsupported_no_trip_id", s.unsupported_no_trip_id_, true);
+  print_if_no_empty("no_vehicle_position", s.no_vehicle_position_, true);
+  print_if_no_empty("vehicle_position_without_position", s.vehicle_position_without_position_,
+                    true);
+  print_if_no_empty("vehicle_position_without_trip", s.vehicle_position_without_trip_, true);
+  print_if_no_empty("vehicle_position_trip_without_trip_id", s.vehicle_position_trip_without_trip_id_,
+                    true);
+  print_if_no_empty("vehicle_position_trip_without_route_id", s.vehicle_position_trip_without_route_id_,
+                    true);
```
felixguendling, line 88: percentages should be relative to vehicles and only vehicles should be relative to total

felixguendling (changes requested):

src/rt/gtfsrt_update.cc
felixguendling: that would make sense - but could also be done in another PR

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +441,149 @@ bool update_run(source_idx_t const src,
...
+                             std::string_view tag,
+                             vector<gtfsrt::FeedEntity> const& entities,
+                             time_point_t const today,
+                             date::sys_seconds message_time,
+                             std::shared_ptr<opentelemetry::trace::Span>& span,
+                             statistics& stats) {
+  // check for duplicate timestamps
+  hash_map<uint64_t, int> counts;
+  for (auto const& ent : entities) {
+    ++counts[ent.vehicle().timestamp()];
```
felixguendling, line 456: has_timestamp?

src/rt/gtfsrt_update.cc
```diff
@@ -463,6 +624,39 @@ statistics gtfsrt_update_msg(timetable const& tt,
...
+        log(log_lvl::error, "rt.gtfs.unsupported",
+            R"(unsupported: no "vehicle_position" field (tag={}, id={}), skipping message)",
+            tag, entity.id());
+        ++stats.no_vehicle_position_;
+      } else if (!entity.vehicle().has_position()) {
+        log(log_lvl::error, "rt.gtfs.unsupported",
+            R"(unsupported: no "position" field in "vehicle_position" field (tag={}, id={}), skipping message)",
+            tag, entity.id());
+        ++stats.vehicle_position_without_position_;
+      } else if (!entity.vehicle().has_timestamp()) {
```
felixguendling, line 639: can't we just assume "now" as timestamp?

src/rt/gtfsrt_update.cc
```diff
@@ -432,6 +592,8 @@ statistics gtfsrt_update_msg(timetable const& tt,
     return {.no_header_ = true};
   }
 
+  vector<gtfsrt::FeedEntity> entities_to_check_;
```
felixguendling, line 595: don't use allocations if not necessary

src/rt/gtfsrt_update.cc
```diff
@@ -586,6 +780,11 @@ statistics gtfsrt_update_msg(timetable const& tt,
     }
   }
 
+  if (use_vp) {
+    handle_vehicle_position(tt, rtt, src, tag, entities_to_check_, today,
```
felixguendling, line 784: handle directly in the loop, don't use allocations if not necessary

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +441,149 @@ bool update_run(source_idx_t const src,
...
+
+      // add rt_transport if not existent
+      if (!r.is_rt()) {
+        r.rt_ = rtt.add_rt_transport(src, tt, r.t_);
+      }
+
+      // match position to stop
+      auto const stop_it = utl::find_if(location_seq, [&](auto const& loc) {
+        auto const stop_position =
+            tt.locations_.get(location_idx_t{loc & 0x0FFFFFFF});
```
felixguendling, line 498: use stop{loc}.location_idx

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +441,149 @@ bool update_run(source_idx_t const src,
...
+        auto const vehicle_position = geo::latlng{vp_lat, vp_lon};
+        return geo::approx_squared_distance(
+                   stop_position.pos_, vehicle_position,
+                   geo::approx_distance_lng_degrees(vehicle_position)) < 10;
```
felixguendling, line 502: there's no point in using this method if approx_distance_lng_degrees is called every time

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +441,149 @@ bool update_run(source_idx_t const src,
...
+      }
+
+      // check for outdated or illegal timestamp
+      auto const vp_ts_utt = unixtime_t{std::chrono::duration_cast<i32_minutes>(
+          std::chrono::seconds{vp.timestamp()})};
+      if ((rtt.rt_transport_latest_vehicle_time_.contains(r.rt_)
+               ? rtt.rt_transport_latest_vehicle_time_[r.rt_] > vp_ts_utt
+               : tt.event_time(r.t_, 0, event_type::kDep) > vp_ts_utt) ||
+          vp_ts_utt > std::chrono::time_point_cast<i32_minutes>(
+                          std::chrono::system_clock::now())) {
```
felixguendling, line 519: performance: don't call `now` all the time - we can assume it's constant for one update run

include/nigiri/rt/rt_timetable.h
```diff
@@ -204,6 +204,12 @@ struct rt_timetable {
   vecvec<rt_transport_idx_t, delta_t> rt_transport_stop_times_;
   vecvec<rt_transport_idx_t, stop::value_type> rt_transport_location_seq_;
 
+  // RT Transport -> latest known VehiclePosition timestamp
+  hash_map<rt_transport_idx_t, unixtime_t> rt_transport_latest_vehicle_time_;
+
+  // RT Transport -> latest known stop through VehiclePosition
+  hash_map<rt_transport_idx_t, stop_idx_t> rt_transport_latest_stop_;
```
felixguendling, line 211: hash_map is too expensive for this - can't afford if not really necessary

include/nigiri/types.h
```diff
@@ -298,6 +298,7 @@ using u8_minutes = std::chrono::duration<std::uint8_t, std::ratio<60>>;
 using duration_t = i16_minutes;
 using unixtime_t = std::chrono::sys_time<i32_minutes>;
 using local_time = date::local_time<i32_minutes>;
+using time_point_t = std::chrono::time_point<date::sys_days::clock, date::sys_days::duration>;
```
felixguendling, line 301: why is this needed? isn't it the same as date::sys_days?

include/nigiri/rt/gtfsrt_update.h
```diff
@@ -35,8 +36,16 @@ struct statistics {
...
   int unsupported_no_trip_id_{0};
+  int no_vehicle_position_{0};
+  int vehicle_position_without_position_{0};
+  int vehicle_position_without_timestamp_{0};
+  int vehicle_position_without_trip_{0};
+  int vehicle_position_trip_without_trip_id_{0};
+  int duplicate_vehicle_position_timestamps_{0};
+  int outdated_or_illegal_vehicle_position_timestamps_{0};
+  int illegal_stop_{0};
+  int vehicle_position_position_not_at_stop_{0};
```
felixguendling, line 48: prefix all with vehicle_position_ if they are only relevant for vehicle position updates

include/nigiri/rt/gtfsrt_update.h
```diff
@@ -48,19 +57,22 @@ statistics gtfsrt_update_msg(timetable const&,
                              rt_timetable&,
                              source_idx_t const,
                              std::string_view tag,
-                             transit_realtime::FeedMessage const&);
+                             transit_realtime::FeedMessage const&,
+                             bool use_vp = false);
```
felixguendling, line 61: `use_vehicle_position` vp = not commonly used

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +441,149 @@ bool update_run(source_idx_t const src,
...
+    } else {
+      ++stats.duplicate_vehicle_position_timestamps_;
+    }
+  }
+  // handle accepted entities
+  for (auto const& entity : accepted_entities) {
+    try {
+      auto const vp = entity.vehicle();
+      auto const vp_lat = vp.position().latitude();
+      auto const vp_lon = vp.position().longitude();
```
felixguendling, line 471: missing check for has_latitude + has_longitude
+ missing test for this check

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +441,149 @@ bool update_run(source_idx_t const src,
...
+      // add rt_transport if not existent
+      if (!r.is_rt()) {
+        r.rt_ = rtt.add_rt_transport(src, tt, r.t_);
+      }
+
+      // match position to stop
+      auto const stop_it = utl::find_if(location_seq, [&](auto const& loc) {
+        auto const stop_position =
+            tt.locations_.get(location_idx_t{loc & 0x0FFFFFFF});
+        auto const vehicle_position = geo::latlng{vp_lat, vp_lon};
```
felixguendling, line 499: build outside loop, store approx_distance_lng_degrees outside loop

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +441,149 @@ bool update_run(source_idx_t const src,
...
+      if (stop_it == end(location_seq)) {
+        log(log_lvl::info, "rt.gtfs.unsupported",
+            R"(Position of Vehicle was not near stop. Skipping Message)", tag,
+            entity.id());
+        ++stats.vehicle_position_position_not_at_stop_;
+        continue;
+      }
+
+      // check for outdated or illegal timestamp
+      auto const vp_ts_utt = unixtime_t{std::chrono::duration_cast<i32_minutes>(
```
felixguendling, line 513: timestamp_case

felixguendling, line 513: utt?

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +441,149 @@ bool update_run(source_idx_t const src,
...
+
+      // get delay
+      auto const et = tt.event_time(r.t_, stopped_at_idx, event_type::kArr);
+      auto const delay = vp_ts_utt - et;
+      auto const delay_cast = std::chrono::duration_cast<duration_t>(delay);
+
+      // update delay for remaining stops
+      auto seq_it = begin(seq_numbers);
+      for (int i = 0; i <= stopped_at_idx; ++i) {
+        ++seq_it;
```
felixguendling, line 554: how is it 100% sure that now the last stop before doesn't have a time >= the first updated stop? can't we check this here and just propagate along the whole trip instead of using large hash_maps that waste RAM?

src/rt/gtfsrt_update.cc
abantobon6: gtfsrt reference states them as required, is that sufficient?

src/rt/gtfsrt_update.cc
abantobon6: happens in `gtfsrt_update_msg(...)` lines: 639 - 642
I could move all the checks to` handle_vehicle_position(...)`?

src/rt/gtfsrt_update.cc
abantobon6: Wouldn't that make it impossible to test for the non-simultaneity of updates in a single message?

felixguendling: we can assume non-simultaneity as anything else wouldn't make any sense

The most important thing is that we still make sure the data model stays in a valid state after every update - no matter if the update made sense or not. This means as long as time series of transports stay increasing, that's fine. No need for extra checks. This can be enforced by always propagating from beginning to end of the transport (!not trip!)

abantobon6: Also "now" would make it very hard to test if the delay is set correctly. I would suggest to use message_time (only) in case of a missing vehicle().timestamp() (or "now", but then there could be that problem again:
> performance: don't call `now` all the time - we can assume it's constant for one update run)

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +432,102 @@ bool update_run(source_idx_t const src,
...
+    auto const seq_numbers = fr.stop_range_;
+
+    // get delay
+    auto const vp_ts = vp.has_timestamp()
+                           ? unixtime_t{std::chrono::duration_cast<i32_minutes>(
+                                 std::chrono::seconds{vp.timestamp()})}
+                           : std::chrono::system_clock::now();
+    auto const et = stopped_at_idx == 0
+                        ? tt.event_time(r.t_, stopped_at_idx, event_type::kDep)
+                        : tt.event_time(r.t_, stopped_at_idx, event_type::kArr);
```
felixguendling, line 503: what happens with additional trips? (`r.is_scheduled() == false`)
those have `r.t_ == transport_idx_t::invalid()`

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +432,102 @@ bool update_run(source_idx_t const src,
...
+      log(log_lvl::info, "rt.gtfs.unsupported",
+          R"(Position of Vehicle was not near stop. Skipping Message)", tag,
+          entity.id());
+      ++stats.vehicle_position_position_not_at_stop_;
+      return;
+    }
+
+    // get remaining stops
+    auto const stopped_at_idx = stop_idx_t(std::distance(begin(location_seq), stop_it));
+    auto const fr = r.is_scheduled() ? frun::from_t(tt, &rtt_const, r.t_)
```
felixguendling, line 492: there's no test for additional trips?
either skip them entirely for now or add a test

(update order matters: first vehicle position or first additional trip with tripUpdate)

src/rt/gtfsrt_update.cc
```diff
@@ -241,7 +254,7 @@ bool add_rt_trip(source_idx_t const src,
     }
     return {};
   };
-  auto const trip_short_name =
+  auto const display_name =
```
felixguendling, line 257: should stay `trip_short_name`

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +432,135 @@ bool update_run(source_idx_t const src,
...
+                             rt_timetable& rtt,
+                             source_idx_t const src,
+                             std::string_view tag,
+                             gtfsrt::FeedEntity const& entity,
+                             date::sys_days const today,
+                             date::sys_seconds const message_time,
+                             std::shared_ptr<opentelemetry::trace::Span>& span,
+                             statistics& stats) {
+
+  // handle accepted entities
```
felixguendling, line 445: not sure what this comment means (just remove)

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +432,135 @@ bool update_run(source_idx_t const src,
...
+    if (!r.is_rt()) {
+      r.rt_ = rtt.add_rt_transport(src, tt, r.t_);
+    }
+
+    // match position to stop
+    auto const vp_position = geo::latlng{vp_lat, vp_lon};
+    auto const app_dist_lng_deg_vp =
+        geo::approx_distance_lng_degrees(vp_position);
+    auto const stop_it = utl::find_if(location_seq, [&](auto const& loc) {
+      auto const stp = tt.locations_.get(stop{loc}.location_idx());
```
felixguendling, line 476: swap stp <-> loc
stp is actually a location
and loc is actually a stop::value_type

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +432,135 @@ bool update_run(source_idx_t const src,
...
+      log(log_lvl::debug, "rt.gtfs.vehicle_update",
+          "Position of Vehicle was not near stop. Skipping message.", tag,
+          entity.id());
+      ++stats.vehicle_position_position_not_at_stop_;
+      return;
+    }
+
+    // get remaining stops
+    auto const stopped_at_idx =
+        stop_idx_t(std::distance(begin(location_seq), stop_it));
```
felixguendling, line 490: always use uniform initialization with curly braces `{...}` to trigger conversion warnings -> add static_cast in this case

src/rt/gtfsrt_update.cc
felixguendling: now is_scheduled is always true - why check?

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +432,135 @@ bool update_run(source_idx_t const src,
...
+    auto const fr = r.is_scheduled() ? frun::from_t(tt, &rtt_const, r.t_)
+                                     : frun::from_rt(tt, &rtt_const, r.rt_);
+    auto const seq_numbers = fr.stop_range_;
+
+    // get delay
+    auto const vp_ts = vp.has_timestamp()
+                           ? unixtime_t{std::chrono::duration_cast<i32_minutes>(
+                                 std::chrono::seconds{vp.timestamp()})}
+                           : std::chrono::system_clock::now();
+    auto const et = stopped_at_idx == 0
```
felixguendling, line 500: ev_time?

felixguendling: et is already "earliest transport"

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +432,135 @@ bool update_run(source_idx_t const src,
...
+      ++stats.vehicle_position_position_not_at_stop_;
+      return;
+    }
+
+    // get remaining stops
+    auto const stopped_at_idx =
+        stop_idx_t(std::distance(begin(location_seq), stop_it));
+    auto const fr = r.is_scheduled() ? frun::from_t(tt, &rtt_const, r.t_)
+                                     : frun::from_rt(tt, &rtt_const, r.rt_);
+    auto const seq_numbers = fr.stop_range_;
```
felixguendling, line 493: this should not be called `seq_numbers`, as it's `const`, just use `fr.stop_range_` everywhere - no alias needed

.pkg.lock
```diff
@@ -24,4 +24,4 @@ opentelemetry-cpp 60770dc9dc63e3543fc87d605b2e88fd53d7a414
 pugixml 60175e80e2f5e97e027ac78f7e14c5acc009ce50
 sol2 40c7cbc7c5cfed1e8c7f1bbe6fcbe23d7a67fc75
 unordered_dense 2c7230ae7f9c30849a5b089fb4a5d11896b45dcf
-wyhash 1e012b57fc2227a9e583a57e2eacb3da99816d99
+wyhash 1e012b57fc2227a9e583a57e2eacb3da99816d99
```
felixguendling, line 27: no need to change this file? no need to commit

src/rt/gtfsrt_update.cc
felixguendling: sequence numbers is a concept from GTFS but those numbers can be arbitrary (9, 99, 999, .. would be valid sequence numbers in stop_times.txt)

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +432,135 @@ bool update_run(source_idx_t const src,
...
+              rtt.unix_event_time(r.rt_, *prev, event_type::kArr) >
+                  rtt.unix_event_time(r.rt_, *prev, event_type::kDep)) {
+            update_delay(tt, rtt, r, *prev, event_type::kArr, delay_cast,
+                         std::nullopt);
+            continue;
+          }
+        }
+        break;
+      }
+    }
```
felixguendling, line 544: this seems overly complex for what it does - can it be simplified for better readability and so it's easier to see if it's correct or not?

why not try something like pairwise(it_range{rbegin, rend})

src/rt/gtfsrt_update.cc
```diff
@@ -419,11 +431,128 @@ bool update_run(source_idx_t const src,
...
+                   std::nullopt);
+      update_delay(tt, rtt, r, second, event_type::kArr, delay_cast,
+                   std::nullopt);
+    }
+
+    // update delay for previous stops if necessary
+    auto const stops_before =
+        interval{fr.stop_range_.from_, stopped_at_idx + 1};
+    for (auto const [curr, prev] :
+         utl::pairwise(it_range(stops_before.rbegin(), stops_before.rend()))) {
```
felixguendling, line 518:
```suggestion
         utl::pairwise(it_range{stops_before.rbegin(), stops_before.rend()})) {
```

PR #293, opened 2025-12-09
Use black or white route text color if contrast is too low

jbruechert:
A good test case is https://api.transitous.org/gtfs/ua_lviv.gtfs.zip

Before:
image

After:
image

The color processing will probably not work correctly on big endian, do we care?

The minimum contrast is taken from [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Guides/Understanding_WCAG/Perceivable/Color_contrast), it might make sense to use the value for icons instead in case this breaks any okay branding.

### DIFF
--- a/src/loader/init_finish.cc
+++ b/src/loader/init_finish.cc
@@ -98,6 +98,53 @@ void assign_importance(timetable& tt) {
   }
 }
 
+// Based on https://www.w3.org/TR/WCAG20/#relativeluminancedef
+float luminance(color_t color) {
+  constexpr auto max = static_cast<float>(std::numeric_limits<uint8_t>::max());
+  auto const r = (color.v_ >> 16 & 0xFF) / max;
+  auto const g = (color.v_ >> 8 & 0xFF) / max;
+  auto const b = (color.v_ & 0xFF) / max;
+
+  auto const color_lum = [](float channel) -> float {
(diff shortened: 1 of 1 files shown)


Findus23:
I am unsure what the best solution is to keep the text readable, but also keep "well-known" color combinations alive.

image

Right now, this turns the text of the U3 in Vienna (https://en.wikipedia.org/wiki/U3_(Vienna_U-Bahn)) black, which feels a bit odd.
But on the other hand according to https://colourcontrast.cc/?background=9c6830&foreground=ffffff the original is a contrast ratio of 2.77 which is far off from the 1:4.5 it should have, so it is indeed quite a readability upgrade.

This is the visual comparison:
image
image


Not sure what the ideal solution is here. Even with U6 (https://en.wikipedia.org/wiki/U6_(Vienna_U-Bahn)) the ["correct" ratio](https://colourcontrast.cc/?background=9c6830&foreground=ffffff) is 4.73, just above 4.5. But flipping the text to black makes things even worse with a ratio of 4.44.
So 4.5 might be a bit to strict of a limit here.


Maybe one solution would be to split the textColor and keep the icon text in the "well-known" color (to make it easier for people to recognize lines at a glance), but if more text is written in a box with colored background a proper high-contrast color is used.

PR #305, opened 2026-01-19
gtfs stop group

felixguendling:


felixguendling:
It's nothing official. No public feed uses it.

PR #309, opened 2026-02-08
Shapes changes for routed shapes

pablohoch:
- Store all segment bounding boxes so that they can be overridden later
- Store shape source (timetable vs. routed)
- Only for routed shapes: Don't add stop positions to the generated polylines


src/rt/frun.cc
```diff
@@ -693,7 +692,12 @@ void frun::for_each_shape_point(
                  get_subshape(common_stops, trip_idx, subrange.from_));
     }
   });
-  consume_pos((*this)[static_cast<stop_idx_t>(range.to_ - 1)].pos(), true);
+  if (consumed == 0U) {
+    consume_pos((*this)[range.from_].pos());
+    consume_pos((*this)[static_cast<stop_idx_t>(range.to_ - 1)].pos());
+  } else if (consumed == 1U && last_pos.has_value()) {
+    consume_pos(*last_pos, true);
+  }
```
MichaelKutzner: If `consumed == 1`, wouldn't that imply `last_pos.has_value()` is true in this case?
I think, I added `from_` and `to_` coordinates to avoid segments with less than `2` points. But this conditions suggests, that this could happen again.

pablohoch:
> @pablohoch @MichaelKutzner looks good to everyone now? :)

Based on some **quick** testing with the AVV data set, it should work with both shapes.txt from the timetable and routed shapes. It seems to look the same as the master version for shapes from the timetable (only tested with AVV).

shapes.txt includes the stop positions:

chrome_2026-02-13_12-51-52

routed shapes don't:

chrome_2026-02-13_12-52-24

PR #333, opened 2026-03-13
Calculation of delays using VehiclePositions and data from past runs

abantobon6:


src/rt/gtfsrt_resolve_run.cc
```diff
@@ -71,4 +74,213 @@ std::pair<run, trip_idx_t> gtfsrt_resolve_run(
...
+      for (auto const route : tt.location_routes_[l]) {
+        auto const location_seq = tt.route_location_seq_[route];
+        for (auto const [stop_idx, s] : utl::enumerate(location_seq)) {
+          if (stop{s}.location_idx() != l) {
+            continue;
+          }
+
+          auto const ev_type =
+              stop_idx == 0 ? event_type::kDep : event_type::kArr;
+
```
abantobon6: language of comments will be English in the next commit

PR #337, opened 2026-03-28
WIP: Export TimeTable to GTFS

PatrickSteil:
Based on the discussion on the [Transitious Matrix](transitous:matrix.spline.de), I decided to implement a simple GTFS Exporter.

@everyone: please give me feedback, review on coding style and so on

I evaluated the exporter on some feeds, and the output was validated by [gtfstidy](https://github.com/patrickbr/gtfstidy) and [gtfs-validator](https://github.com/MobilityData/gtfs-validator) (that's far from saying the exporter works correctly, but on these instances the feed was valid).

TODOs:
- [ ] export as zip directly, maybe toggle by CLI flag
- [ ] make the calendar.txt export "easier" to understand (prev i wrote everything into calendar_dates.txt, seemed kind of ugly)
- [ ] find wheelchair accessibility for trips.txt


exe/export.cc
```diff
@@ -0,0 +1,44 @@
...
+
+using namespace nigiri;
+
+int main(int argc, char* argv[]) {
+  namespace bpo = boost::program_options;
+
+  auto tt_path = std::filesystem::path{};
+  auto gtfs_dir = std::filesystem::path{};
+
+  bpo::options_description desc("Allowed options");
```
felixguendling:
```suggestion
  auto desc = bpo::options_description{"Allowed options"};
```

Please transform every variable declaration into `auto [const] name = ctor{...};` (note the curly braces for uniform initialization, make everything `auto const` if possible, otherwise `auto`) so we keep it consistent. I know there's old cold not in this style, but we try to have new code in [AAA style](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/).

src/gtfs_export/export_gtfs.cc
```diff
@@ -0,0 +1,410 @@
...
+
+  if (!needs_quoting || input.empty()) {
+    return std::string{input};
+  }
+
+  auto out = std::string{};
+  out.reserve(input.size() + 2);
+  out.push_back('"');
+  for (auto const c : input) {
+    if (c == '"')
```
felixguendling: could you please add curly braces everywhere to each if/loop statement? :)

PR #345, opened 2026-04-15
Advanced transfer times

aff3npirat:
Correctly consider `additionalTransferTime`, `minTransferTime`, `transferTimeFactor` when routing.

Fixes motis-project/motis#1357

### DIFF
--- a/src/routing/raptor/reconstruct.cc
+++ b/src/routing/raptor/reconstruct.cc
@@ -497,33 +497,6 @@ void reconstruct_journey_with_vias(timetable const& tt,
           v + 1, v, stay_fp_target);
     }
 
-    // if (v != 0 && matches(tt, location_match_mode::kEquivalent,
-    //                       q.via_stops_[v - 1].location_, l)) {
-    //   --v;
-    //   if (matches(tt, location_match_mode::kEquivalent, l, fp.target())) {
-    //     stay_fp_target = q.via_stops_[v].stay_;
-    //     trace_reconstruct(
-    //         "  [check_fp]: fp start+target matches current via: v={}->{}, "
-    //         "stay_target={}\n",
-    //         v + 1, v, stay_fp_target);
-    //   } else {
-    //     stay_l = q.via_stops_[v].stay_;
-    //     trace_reconstruct(
-    //         "  [check_fp]: fp start matches current via: v={}->{},
-    //         stay_l={}\n", v + 1, v, stay_l);
-    //   }
-    // } else if (v != 0 && matches(tt, location_match_mode::kEquivalent,
-    //                              q.via_stops_[v - 1].location_, fp.target()))
-    //                              {
-    //   --v;
-    //   assert(stay_fp_target == 0_minutes);
-    //   stay_fp_target = q.via_stops_[v].stay_;
-    //   trace_reconstruct(
-    //       "  [check_fp]: fp target matches current via: v={}->{}, "
-    //       "stay_fp_target={}\n",
-    //       v + 1, v, stay_fp_target);
-    // }
-
     auto const fp_plus_stay_l_duration = fp_duration + stay_l.count();
     auto const fp_plus_both_stay_duration =
         fp_duration + stay_l.count() + stay_fp_target.count();
--- a/src/routing/raptor/reconstruct.cc
+++ b/src/routing/raptor/reconstruct.cc
@@ -661,7 +661,7 @@ void reconstruct_journey_with_vias(timetable const& tt,
       auto const footpaths = kFwd ? tt.locations_.footpaths_in_[q.prf_idx_][l]
                                   : tt.locations_.footpaths_out_[q.prf_idx_][l];
       for (auto const& fp : footpaths) {
-        auto fp_legs = check_fp(k, l, curr_time, fp, false, true);
+        auto fp_legs = check_fp(k, l, curr_time, fp, true, true);
         if (fp_legs.has_value()) {
           return std::move(*fp_legs);
         }
--- a/src/routing/raptor/reconstruct.cc
+++ b/src/routing/raptor/reconstruct.cc
@@ -464,18 +464,11 @@ void reconstruct_journey_with_vias(timetable const& tt,
 
     auto const backup_v = v;
 
-    auto const is_intermodal =
-        q.dest_match_mode_ == location_match_mode::kIntermodal;
     auto stay_l = 0_minutes;
     auto stay_fp_target = 0_minutes;
     trace_reconstruct("  [check_fp] v={}, l={}, fp.target={}, intermodal={}\n",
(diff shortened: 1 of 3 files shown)


aff3npirat:
Also I noticed that in the journey details, footpaths always state the full transfer time instead of estimated walk time e.g. `32 min Walk (78m)`. Is this intended even for advanced transfer time settings?

@felixguendling @traines-source

felixguendling:
@pablohoch

pablohoch:
> @pablohoch

The last two arguments to `check_fp` are `bool adjust_transfer_time` and `bool is_td_footpath`. I'm not sure why `adjust_transfer_time` was `false` here. Looks like it was changed [here](https://github.com/motis-project/nigiri/commit/d17caf7b6e5bf3b0f9693ef8ba03b7aa5bb5e557#diff-0f61b9c0ff6b7f3362b5e95c6331b8cf5932af1e526b33fcd2e2619de322bde9L435) (if the link doesn't work - reconstruct.cc, line 435 before, 502 after) - it was `true` before (maybe a mix-up of the two booleans?). Unless there was a reason to change it that I'm unaware of, it should be correct to change it back to `true` via this PR.
There is another occurrence below that (currently line 682), which should probably also be changed. Also in both cases (line 664 + 682) `is_td_footpath` is `true`, which seems wrong as only the second case is for td footpaths.

So to me it looks like the correct way should be:
Line 664: `auto fp_legs = check_fp(k, l, curr_time, fp, true, false);` (adjust = true, td = false)
Line 682: `auto fp_legs = check_fp(k, l, curr_time, fp, true, true);` (adjust = true, td = true)
(Haven't looked at it in detail though, but based on how it was originally + the naming, this makes the most sense.)

> Also I noticed that in the journey details, footpaths always state the full transfer time instead of estimated walk time e.g. 32 min Walk (78m). Is this intended even for advanced transfer time settings?

Sounds like a bug to me.

felixguendling:
Looks like I refactored it and introduced a bug. Probably an indicator that we need more tests.

@aff3npirat can you please adapt it like @pablohoch proposed? Maybe you can add a test case that fails in the current master state and gets green with your fix?

aff3npirat:
Okay I will look into the failed tests.

> Sounds like a bug to me.

I think the problem is that footpaths always have adjusted transfer times as duration and their arrival time is always set to `depTime + adjusted_duration`. Changing to walking time (adjusted by walking speed) as duration and setting arrival time accordingly would fix this issue.

src/routing/raptor/reconstruct.cc
```diff
@@ -488,15 +497,40 @@ void reconstruct_journey_with_vias(timetable const& tt,
                           q.via_stops_[v - 1].location_, fp.target())) {
       --v;
       assert(stay_fp_target == 0_minutes);
-      if (is_intermodal) {
```
aff3npirat, line 492: This caused the failed tests, by searching for transports with the wrong time. Previously this was  "fixed" by treating fp as td (looking for transports with `better_or_equal` instead of exact event times).

From my understanding intermodal fps are only used for first/final leg in journeys, so I dont quite understand why this `if` was here in the first place.

aff3npirat:
So the test cases are fixed. I would like to keep this PR open and start working on fixing the fp durations so that they actually reflect the walking time and not transfer time. @felixguendling

src/routing/raptor/reconstruct.cc
felixguendling: @pablohoch do you understand this?
I'm a bit lost

felixguendling:
> So the test cases are fixed. I would like to keep this PR open and start working on fixing the fp durations so that they actually reflect the walking time and not transfer time. @felixguendling

Makes sense.

src/routing/raptor/reconstruct.cc
pablohoch: I'm not entirely sure what failed tests this refers to - I assume this is related to the other changes that cause some tests to fail?

That `if` was introduced in https://github.com/motis-project/nigiri/pull/158, but back then it was `if (!is_final_leg || is_intermodal)`. This was part of a fix for the second bug listed in the PR - journeys where the last stop is a via stop with a stay duration.
The idea was to ignore the stay duration at the final stop if that stop is a station, because it can't be properly represented in the output (since no intermodal footpath is added after it) - that's what the if condition originally checked.

Then it looks like the if condition was changed from `if (!is_final_leg || is_intermodal)` to the current `if (is_intermodal)` in https://github.com/motis-project/nigiri/pull/296, which is also related to td footpaths. I'm not sure about that but it could be related to the `check_fp` refactoring bug, or maybe there is another reason (I'm not really familiar with the td footpath implementation).
Notably that PR also changed the `via_test_31_M_Q_via_Q_10m` test from `[2019-05-01 09:00, 2019-05-01 10:00]` to `[2019-05-01 09:00, 2019-05-01 10:10]`, while the last arrival at stop Q is still `10:00`.
So it looks like in that PR the handling of stay duration at a final station stop changed to include it in the journey duration (there are a lot of other changes as well, maybe this is required or more consistent with td footpaths - I don't know).

While I still think it makes sense to ignore the via stay duration at a final **station** destination, it's probably also okay to include it, as the whole scenario doesn't make too much sense IMHO. Either we ignore the requested stay duration or we include it in the journey duration but not in the legs (because we can't without adding another footpath).

Therefore I think we could probably remove the `if (is_intermodal)` check (and always apply the stay duration) if that behavior is desired (journey ends at 10:10 even though the final arrival at the destination station is at 10:00). Alternatively, try changing the if statement back to the original ` if (!is_final_leg || is_intermodal)` and adjust the `via_test_31_M_Q_via_Q_10m` back to expect a journey end at 10:00 and see if that now works with the `check_fp` fixes (I find this a bit less confusing).

aff3npirat: > try changing the if statement back to the original  if (!is_final_leg || is_intermodal) and adjust the via_test_31_M_Q_via_Q_10m back to expect a journey end at 10:00 and see if that now works with the check_fp fixes

Still adds stay duration because `raptor` was also changed to include stay duration (`raptor.h, line 461`, introduced in #296). Changing this and `pong` throws for `via_test_31_M_Q_via_Q_10m`. I think I'll just leave it as is for now and work on the footpath lenghts first.

felixguendling (approved):

traines-source:
What's the status here? The fixed transfer times look good. Regarding the reported walking times, they just include the additional/min/factor, they do not cover the entire time between the transit legs. From my perspective this could be considered expected behaviour. At least for transferTimeFactor, I think you would want to consider it as part of the walking time (because it means that you want to walk slowlier). Then for consistency I would argue it's easier to also include additonal and minTransferTime in the walking time. In any case I would propose merging this because this is quite an ugly bug that now is finally fixed :) Then we can still think about including it into walking times or not.

felixguendling:
PR has conflicts

PR #347, opened 2026-04-15
Fix multiple definitions of trace macro

### LINKED ISSUE #346: Compiling with debug logs fails
When either of `NIGIRI_DIJKSTRA_TRACING` or `NIGIR_TRACING` is defined compilation fails because `trace` macro is defined multiple times (by `routing/dijkstra.cc` and `routing/raptor/debug.h`.

aff3npirat:
Fixes #346

### DIFF
--- a/include/nigiri/common/delta_t.h
+++ b/include/nigiri/common/delta_t.h
@@ -1,7 +1,8 @@
 #pragma once
 
-#include "nigiri/types.h"
 #include "fmt/base.h"
+
+#include "nigiri/types.h"
 
 namespace nigiri {
 
--- a/include/nigiri/common/delta_t.h
+++ b/include/nigiri/common/delta_t.h
@@ -1,7 +1,7 @@
 #pragma once
 
-#include "nigiri/routing/raptor/debug.h"
 #include "nigiri/types.h"
+#include "fmt/base.h"
 
 namespace nigiri {
 
@@ -17,10 +17,10 @@ template <typename T>
 inline delta_t clamp(T t) {
 #if defined(NIGIRI_TRACING)
   if (t < std::numeric_limits<delta_t>::min()) {
-    trace_upd("CLAMP {} TO {}\n", t, std::numeric_limits<delta_t>::min());
+    fmt::print("CLAMP {} TO {}\n", t, std::numeric_limits<delta_t>::min());
   }
   if (t > std::numeric_limits<delta_t>::max()) {
-    trace_upd("CLAMP {} TO {}\n", t, std::numeric_limits<delta_t>::max());
+    fmt::print("CLAMP {} TO {}\n", t, std::numeric_limits<delta_t>::max());
   }
 #endif
 
--- a/src/loader/gtfs/seated.cc
+++ b/src/loader/gtfs/seated.cc
@@ -4,12 +4,8 @@
 
 #include <ranges>
 
-// #define NIGIRI_SEATED_DEBUG
-#ifdef NIGIRI_SEATED_DEBUG
-#define trace(...) fmt::println(__VA_ARGS__)
-#else
 #define trace(...)
-#endif
+// #define trace(...) fmt::println(std::clog, __VA_ARGS__)
 
 namespace sv = std::views;
 
--- a/src/loader/gtfs/seated.cc
+++ b/src/loader/gtfs/seated.cc
@@ -4,8 +4,12 @@
 
 #include <ranges>
 
-// #define trace(...) fmt::println(std::clog, __VA_ARGS__)
+// #define NIGIRI_SEATED_DEBUG
+#ifdef NIGIRI_SEATED_DEBUG
+#define trace(...) fmt::println(__VA_ARGS__)
+#else
 #define trace(...)
+#endif
 
 namespace sv = std::views;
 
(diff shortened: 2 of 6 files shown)


include/nigiri/routing/raptor/debug.h
```diff
@@ -11,9 +11,9 @@
 #define trace_upd(...) fmt::print(__VA_ARGS__)
 
 #ifdef NIGIRI_RAPTOR_TRACING_ONLY_UPDATES
-#define trace(...)
+#define trace_raptor(...)
```
felixguendling: this can stay `trace` if `trace_dijkstra` gets changed?

include/nigiri/routing/raptor/raptor.h
```diff
@@ -292,7 +292,7 @@ struct raptor {
       for (auto k = 1U; k != end_k; ++k) {
         auto const dest_time = round_times_[k][i][Vias];
         if (dest_time != kInvalid) {
-          trace("ADDING JOURNEY: start={}, dest={} @ {}, transfers={}\n",
+          trace_raptor("ADDING JOURNEY: start={}, dest={} @ {}, transfers={}\n",
```
felixguendling: please enable format-on-save

aff3npirat: I will, sorry that it happend so often...

include/nigiri/routing/raptor/debug.h
aff3npirat: Yes it can. I renamed it to avoid problems with other dependencies/projects using nigiri.
Should I rename it back?

felixguendling: we can #undef the macro at the end of each header file using it
but actually this macro should ideally not leak as it's in an implementation file
maybe we can reorganize the code so other code does not depend on implementation details of the nigiri library

for now keep it `trace`

aff3npirat: Ok will rename it.
I also thought of `#undef` but e.g. `trace_raptor` is used by other header files where it is not defined - so we would have to redefine the macro. I think if we can solve this issue by restructuring it would be much cleaner, and easier to keep track where each macro is used.

aff3npirat:
The problem was an unused import of `rt_timetable.h (-> delta_t.h)`.
I removed the import and `delta_t` no longer depends on `debug.h` + added macros to enable/disable logging for other parts.

include/nigiri/common/delta_t.h
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
-#include "nigiri/routing/raptor/debug.h"
 #include "nigiri/types.h"
+#include "fmt/base.h"
```
felixguendling, line 3: header ordering (see STYLE.md)

src/loader/gtfs/seated.cc
```diff
@@ -4,8 +4,12 @@
 
 #include <ranges>
 
-// #define trace(...) fmt::println(std::clog, __VA_ARGS__)
+// #define NIGIRI_SEATED_DEBUG
+#ifdef NIGIRI_SEATED_DEBUG
+#define trace(...) fmt::println(__VA_ARGS__)
+#else
 #define trace(...)
+#endif
```
felixguendling: what's wrong with having the simple one line solution?

src/routing/ontrip_train.cc
```diff
@@ -5,16 +5,14 @@
 #include "nigiri/routing/query.h"
 #include "nigiri/timetable.h"
 
-namespace nigiri::routing {
-
-constexpr auto const kTracing = true;
+// #define NIGIRI_ONTRIP_DEBUG
```
felixguendling: couldn't this also be `// #define trace(...) fmt::println(std::clog, __VA_ARGS__)`

src/loader/gtfs/seated.cc
aff3npirat: I just thought it is more convenient to enable/disable logging by passing compiler flags than actually changing source code. This prevents forgetting to disable logging before commiting/pushing.

src/routing/ontrip_train.cc
aff3npirat: Yes, same reasoning for my implementation as in reply to
> what's wrong with having the simple one line solution?

src/loader/gtfs/seated.cc
felixguendling: I don't think that's the approach people take. At least not me. Uncommenting one line of code is the most simple solution I can think of and not checking this change into Git is doable.

aff3npirat: Okay will do. Should I also change `fares.cc` to use a simple comment/uncomment approach (and maybe also `dijkstra.cc`) or simply leave everthing as is?

felixguendling: > Should I also change fares.cc to use a simple comment/uncomment approach (and maybe also dijkstra.cc)

yes

aff3npirat: Done. `fares.cc` couldnt be changed this way because `NIGIRI_FARES_DEBUG` macro actually changes code instead of only handling `trace` definition (see line 358).

felixguendling: let me know when it's ready for review

PR #351, opened 2026-05-07
Add coupled trips support for pong

JulienDidier38:
Add a new query parameter and associated business logic to enable the retrieval of coupled trips.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /.clang-tidy
 /.vscode
 shape*.bin
+.DS_Store
--- a/include/nigiri/abi.h
+++ b/include/nigiri/abi.h
@@ -67,6 +67,16 @@ struct nigiri_event_change {
 };
 typedef struct nigiri_event_change nigiri_event_change_t;
 
+// An alternative transport that covers the same leg with identical dep/arr
+// times (e.g. Flügelzug / coupled train on a parallel route).
+struct nigiri_leg_alternative {
+  uint32_t transport_idx;
+  uint16_t day_idx;
+  uint16_t enter_stop_idx;
+  uint16_t exit_stop_idx;
+};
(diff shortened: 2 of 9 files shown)


felixguendling:
Maybe we should add a `CONTRIBUTING.md` file or similar to state that solutions should be discussed before being worked on. I think this approach is not really good. First, it solves it only for PONG but for results paging (earlier forward) and (later backward) we use rRAPTOR, so results would actually be inconsistent. Second, we have now 4 places with similar functionality that should be one generic function called with different parameters:

- itinerary id
- slowDirect
- pong.cc get_earliest_alternative
- and now this
- and even this https://github.com/motis-project/motis/issues/546 will require the same functionality

So instead of adding separate journeys to the result set (which adds a lot of redundant data), it might make sense to add alternatives for each leg separately that would still lead to a feasible journey like it's suggested here https://github.com/motis-project/motis/issues/546. This solves the same problem (client can still decide to display those as separate journeys) but in a way more generic way.

JulienDidier38:
Hi Felix, 
I was about to post a message on the Matrix channel, but you were too quick. 
This work is a proof of concept to reassure our clients about Motis's ability to address all our use cases. Managing coupled journeys is a major challenge for SNCF Connect and could be a stumbling block for integrating the solution into the SNCF Connect ecosystem. 
We tried the GTFS adaptations suggested in the Matrix channel, but without success. We'll look at the link you mentioned and see how we can adapt our proposal.

felixguendling:
I have already something working here:
https://github.com/motis-project/motis/pull/1398

You can try (it's still work-in-progress but should basically work) it. Let me know if that works for you.

JulienDidier38:
Hi Felix, I'm already testing this. It seems to fit our needs. In any case, we can clearly see our twin trains rising in the alternatives.

PR #356, opened 2026-05-17
Implement Google Transit Ticketing Extension

jbruechert:


include/nigiri/timetable.h
```diff
@@ -463,6 +463,22 @@ struct timetable {
   string_store<language_idx_t> languages_;
 
   cista::base_t<source_idx_t> n_sources_{};
+
+  // Ticketing
+  hash_map<provider_idx_t, hash_map<location_idx_t, string>>
```
felixguendling: not sure if using so many hashmaps won't backfire on transitous scale?
maybe I should implement a sorted_vector_map in cista or something similar so we can have a compact representation for sparse data?

jbruechert: With the current level of ticketing support it is unlikely to become an issue, but something like that sounds nice.
Generally, please let me know if there is a better data structure for what I'm doing, I do not feel like I have a full overview of the cista features yet :)

include/nigiri/timetable.h
```diff
@@ -487,6 +488,28 @@ struct timetable {
...
+
+  struct ticketing_link {
+    ticketing_link_type type_;
+    string web_;
+    string android_;
+    string ios_;
+  };
+
+  vecvec<ticketing_link_idx_t, ticketing_link> ticketing_links_;
+  hash_map<provider_idx_t, ticketing_link_idx_t> ticketing_agencies_;
```
felixguendling: this can go into provider struct directly?

include/nigiri/timetable.h
```diff
@@ -487,6 +488,28 @@ struct timetable {
...
+  hash_map<route_id_idx_t, ticketing_link_idx_t> ticketing_routes_;
```
felixguendling: this could go into the route_id storage directly?

include/nigiri/timetable.h
```diff
@@ -487,6 +488,28 @@ struct timetable {
...
+  enum class ticketing_link_type : uint8_t {
+    kGoogleTransit,
+  };
+
+  struct ticketing_link {
+    ticketing_link_type type_;
+    string web_;
+    string android_;
+    string ios_;
+  };
```
felixguendling: strings usually carry overhead
why not data-oriented programming like the rest?

something in this direction?

vector_map<ticketing_link_idx_t, ticketing_link_type> type_
vecvec<ticketing_link_idx_t, char> web_
vecvec<ticketing_link_idx_t, char> andoid_
vecvec<ticketing_link_idx_t, char> ios_

include/nigiri/timetable.h
```diff
@@ -487,6 +488,28 @@ struct timetable {
...
+
+  // Ticketing
+  hash_map<provider_idx_t, hash_map<location_idx_t, string>>
+      location_ticketing_identifier_;
+  hash_map<trip_idx_t, string> trip_ticketing_identifier_;
+  hash_map<location_idx_t, std::monostate> location_ticketing_unavailable_;
+  hash_map<trip_idx_t, std::monostate> trip_ticketing_unavailable_;
+
+  enum class ticketing_link_type : uint8_t {
+    kGoogleTransit,
```
felixguendling: if there's only one - does it make sense to have this?

we can always change the data model - it's not like the API that we want to keep it stable

now we waste bytes because all will carry the same information?

include/nigiri/timetable.h
```diff
@@ -487,6 +488,28 @@ struct timetable {
   string_store<language_idx_t> languages_;
 
   cista::base_t<source_idx_t> n_sources_{};
+
+  // Ticketing
+  hash_map<provider_idx_t, hash_map<location_idx_t, string>>
+      location_ticketing_identifier_;
+  hash_map<trip_idx_t, string> trip_ticketing_identifier_;
+  hash_map<location_idx_t, std::monostate> location_ticketing_unavailable_;
```
felixguendling: bitvec_map<location_idx_t> in locations maybe?

include/nigiri/timetable.h
```diff
@@ -487,6 +488,28 @@ struct timetable {
...
+  hash_map<trip_idx_t, std::monostate> trip_ticketing_unavailable_;
```
felixguendling: bitvec_map<trip_idx_t> maybe?

include/nigiri/timetable.h
```diff
@@ -487,6 +488,28 @@ struct timetable {
   string_store<language_idx_t> languages_;
 
   cista::base_t<source_idx_t> n_sources_{};
+
+  // Ticketing
+  hash_map<provider_idx_t, hash_map<location_idx_t, string>>
```
felixguendling: vecvec<location_idx_t, pair<provider_idx_t, string_idx_t>> maybe?

include/nigiri/loader/gtfs/agency.h
```diff
@@ -14,13 +14,15 @@ struct timetable;
 namespace nigiri::loader::gtfs {
 
 using agency_map_t = hash_map<std::string, provider_idx_t>;
+using agency_ticketing_map_t = hash_map<provider_idx_t, string>;
```
felixguendling: could be `std::string`?

src/loader/gtfs/ticketing.cc
```diff
@@ -0,0 +1,130 @@
...
+#include "nigiri/loader/gtfs/trip.h"
+#include "nigiri/timetable.h"
+
+#include "utl/parser/csv_range.h"
+#include "utl/progress_tracker.h"
+
+constexpr auto kTicketingDeeplinks = "ticketing_deep_links.txt";
+constexpr auto kTicketingIdentifiers = "ticketing_identifiers.txt";
+
+namespace nigiri::loader::gtfs {
```
felixguendling, line 15:
```suggestion
namespace nigiri::loader::gtfs {

```

src/loader/gtfs/ticketing.cc
```diff
@@ -0,0 +1,130 @@
...
+    timetable& tt, std::string_view file_content) {
+  struct ticketing_deep_links_row {
+    utl::csv_col<utl::cstr, UTL_NAME("ticketing_deep_link_id")>
+        ticketing_deep_link_id;
+    utl::csv_col<utl::cstr, UTL_NAME("web_url")> web_url;
+    utl::csv_col<utl::cstr, UTL_NAME("android_intent_uri")> android_intent_uri;
+    utl::csv_col<utl::cstr, UTL_NAME("ios_universal_link_url")> ios_url;
+  };
+
+  hash_map<std::string_view, ticketing_link_idx_t> map;
```
felixguendling:
```suggestion
  auto map = hash_map<std::string_view, ticketing_link_idx_t>{};
```

felixguendling (commented):
in general good work!

I'm a bit worried about scalability.

Until now we had pretty good success with "not running into scalability issues" (which can quickly happen on Transitous scale) by consistently using data-oriented programming as much as possible. This PR rarely uses data-oriented programming. Do you see any downsides when changing it? I understand that it's expected that data (-> also vectors) would be sparse at the beginning. But I would see advantages to not have to care about if it stays like this. That's why I proposed some changes in this direction. But I'm open for discussion (that's why my comments are more a "maybe .." than "this has to change").

Edit: to clarify: I see `hash_map` as a waste of space if it's not really necessary for speed (e.g. for stop ids) and `string` always has 24 byte for (pointer, size, capacity), so vecvec is more optimal (less waste). Of course for sparse data, you have lot of empty entries unfortunately..

include/nigiri/timetable.h
jbruechert: It felt wrong to hardcode input specific logic in the API layer, but the enum is indeed not currently necessary.

PR #361, opened 2026-05-29
Fix via stay duration

aff3npirat:


traines-source:
LGTM, then we can remove the rRAPTOR fork workaround in motis?

PR #366, opened 2026-06-12
Fix pong via stays

aff3npirat:


traines-source:
The msvc build should now hopefully be fixed by merging master

PR #367, opened 2026-06-14
Adding trips.txt > trip_route_type support (MBTA GTFS spec)

orhazal:
Following the discussions in https://github.com/google/transit/issues/358 and https://github.com/google/transit/issues/409 : the SNCF GTFS feed has routes whose individual trips use different transport modes (e.g. a rail route where some trips run as replacement buses). GTFS only carries route_type at the route level, so these trips can't be classified or filtered correctly.

This PR adds optional support for the MBTA convention: a trip_route_type column in trips.txt that overrides routes.txt route_type for a single trip when set.

Implementation: the per-trip type feeds into the existing route-key grouping in load_timetable, so an overridden trip naturally lands in its own route with the correct clasz. No new per-trip storage in the timetable, it only steers which route a trip groups under. A test is included.

Fully backward-compatible: feeds without the column (or with empty values) load exactly as before.

@felixguendling : you were part of those discussions; would you be open to this in nigiri? Happy to adjust the approach. 
It will enable proper filtering on a modified/patched SNCF GTFS feed. Thank you.


traines-source:
I think I personally am part of the "route duplication" faction (be it for rail replacement or mixed TER service, where for both IMO it is sufficient to just give the same or similar name to two distinct routes to make it clear to the user that it is essentially the same).

But still, since this is quite a straightforward change, if you think it is useful to you, I have no objections to merge it.

I think there are a few places where we rely on info stored with the route_ids (i.e. on the GTFS routes and not the nigiri routes). ServiceAlerts and GTFS-RT ADDED trips come to my mind, and the `routeType` field in the API response (meaning while the `mode` field would be correct, the `routeType` field would contain the original type of the route for these special trips if I'm not mistaken – haven't tested it). There might be more places.

orhazal:
@traines-source I agree with you. For me too, the "cleanest" solution would be to duplicate the routes. I'm thinking in particular of the IDFM GTFS feed, which treats RER replacement buses as separate "routes" that share the name (and other attributes) of the main route.

But before proposing this PR, I first raised the [question](https://www.data.gouv.fr/datasets/horaires-sncf/discussions?discussion_id=69f20ea848f70de09aecc54c) with the producer of the SNCF GTFS feed, who told me that it's due to their internal machinery / pipelines (passenger information / commercial consistency) and that they couldn't do otherwise. Especially since the GTFS generation happens at the very end of the chain.

I also looked into the GTFS spec to see whether there were similar concerns, and that's how i found issues 358 and 409 in google/transit. But given the fact that it hasn't been discussed since last summer, I don't think it'll be addressed any time soon, hence this PR.

orhazal:
Note : the build error is caused by rtree.c dependency and is present in master too (?)

traines-source:
> Note : the build error is caused by rtree.c dependency and is present in master too (?)

Yes not sure where that comes from, might be due to an update of msvc. I haven't come around looking at it yet.

> I first raised the [question](https://www.data.gouv.fr/datasets/horaires-sncf/discussions?discussion_id=69f20ea848f70de09aecc54c) with the producer of the SNCF GTFS feed

Thanks for the hint, I might finally ask two questions there that have been bothering me for a long time (extended route types and RT platform information)

orhazal:
@traines-source For the platform information, I ran into the same problem last year while fiddling with the Transilien / RER feed (IDFM) and their monomodalStopPlace modelisation (stop with type=0 but is equivalent to the "station" or the entirety of the Transilien platforms within a station),

The answer I got from Transilien was that operating/circulation rules change, sometimes at the last minute, and to avoid surfacing incorrect information and to stay consistent, the decision was made to expose the monomodalStopPlace rather than the platform as the point of passage in the GTFS.
I guess it's pretty much the same problem here. I've never seen train stations modeled by platform in the French feeds, and I think it's also due to the fact that platform assignments can change at the last moment.

However, the info exists in the NeTEx feed but i'm not quite sure if it's "plugged" (the stop points are monomodalStopPlaces too)

Here is an example of Châtelet les Halles in Paris
image

The RER D can use quays 3 or 4 depending of the circulation plan. So all stop points for the RER in Châtelet les Halles are represented by the id IDFM:monomodalStopPlace:45102

traines-source:
I did not have the intention to topple the tradition of "platform is announced 20 minutes prior to departure" :D As I have now described over in that forum, I just propose to make it part of the GTFS-RT basically as a platform change, and in order to do that, introduce "dummy" platforms in stops.txt that are just not used in the static GTFS.

orhazal:
@traines-source thanks for the info ! Makes a lot of sense indeed.
If they can implement the extended types + this, it will be great for all consumers !

traines-source:
The msvc build should now hopefully be fixed by merging master

orhazal:
> But still, since this is quite a straightforward change, if you think it is useful to you, I have no objections to merge it.

Is it still okay for you? The build is OK. Thank you

traines-source:
If you're ok with the caveats I mentioned :)

PR #376, opened 2026-07-05
partially implemented McRaptor algorithm

HowToChangeMyName:
added McRaptor as a new file (mcraptor.h). to determine which legs are discarded a new struct (bag) is used which can store multiple variables and are easier to extend.

TODO:
- so far only tmp_ and best_ ( should be replaced with new_tmp_/new_best_ and renamed)
- constructor/member creates copies of data which should be optimized
- to switch between raptor and raptor, there is a macro in pong.cc. for now, it is not possible to switch algorithm in the executable

### DIFF
--- a/include/nigiri/routing/raptor/mcraptor.h
+++ b/include/nigiri/routing/raptor/mcraptor.h
@@ -1,0 +1,1412 @@
+#pragma once
+
+#include <cassert>
+
+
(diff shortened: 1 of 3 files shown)


felixguendling:
too far from anything that could resemble a McRAPTOR - closing

PR #379, opened 2026-07-11
use values from files as feed_info.default_language translation

felixguendling:


jbruechert:
Wouldn't the strings from individual files correspond to `feed_info.feed_lang` instead of `feed_info.default_lang`?

felixguendling:
Yes, probably that's better.

PR #381, opened 2026-07-14
skip non-existing rt stops without sequence number

### LINKED ISSUE #1508: GTFS RT update: check if stop even exists while scanning
https://github.com/public-transport/transitous/issues/2262#issuecomment-4868354919

traines-source:
closes https://github.com/motis-project/motis/issues/1508

### DIFF
--- a/src/rt/gtfsrt_update.cc
+++ b/src/rt/gtfsrt_update.cc
@@ -351,29 +351,26 @@ bool update_run(source_idx_t const src,
     auto const loc_idx = stop{location_seq[stop_idx]}.location_idx();
     auto matches = false;
     while (upd_it != end(stus)) {
-      auto matches_seq = r.is_scheduled() && upd_it->has_stop_sequence() &&
-                         upd_it->stop_sequence() == *seq_it;
+      matches = (r.is_scheduled() && upd_it->has_stop_sequence() &&
+                 upd_it->stop_sequence() == *seq_it) ||
+                upd_it->stop_id() == tt.locations_.ids_[loc_idx].view();
+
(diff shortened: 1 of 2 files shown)


src/rt/gtfsrt_update.cc
```diff
@@ -349,12 +349,32 @@ bool update_run(source_idx_t const src,
...
-         (upd_it->has_stop_id() &&
-          upd_it->stop_id() == tt.locations_.ids_[loc_idx].view()));
+    auto matches = false;
+    while (upd_it != end(stus)) {
+      matches = (r.is_scheduled() && upd_it->has_stop_sequence() &&
+                 upd_it->stop_sequence() == *seq_it) ||
+                (upd_it->has_stop_id() &&
+                 upd_it->stop_id() == tt.locations_.ids_[loc_idx].view());
+
+      if (matches || upd_it->has_stop_sequence() || !upd_it->has_stop_id()) {
```
felixguendling: Stop update having a stop sequence number is no guarantee that we can find this stop? If we don't have a matching stop sequence number, it's probably still better to skip this as it will have the same problem like the "stop id not existent" bug we want to solve.

My idea was to just check exactly the `matches` predicate for this and all following stops. If there's one -> fine, if not: skip (as we will not find a match and then not process the rest).

felixguendling:
Maybe we should have a test case for this, so nobody will destroy it? Or will this be added in MOTIS?

src/rt/gtfsrt_update.cc
traines-source: the main goal was to preserve support for platform changes. but yeah true we might still be able to check for both

traines-source:
I had added a test case for platform change and for the unknown stop id case

PR #382, opened 2026-07-14
Wrap access to lower bound in helper methods

PatrickSteil:
As discussed with Felix, if the lower bound vector is empty, we access it somewhere without any checks. Also, when disabling `kUseLowerBounds`, we return the valid lower bound of 0


felixguendling:
Nice, thank you! :)

PR #387, opened 2026-07-24
assume sorted offsets for performance

traines-source:
not sure why in rraptor we sort offsets and in pong we don't (?). but I propose to always sort them over in motis.

The lookup_offset loop is becoming quite expensive in certain cases (dense city with lots of direct routes and lots of offsets). Even with this tweak it can still consume 3s for the default `numLegAlternatives=3` (e.g. https://test.motis-project.org/?time=2026-07-22T10%3A54%3A00.000Z&fromPlace=48.84103907668603%2C2.316604555818344&toPlace=48.87615362084247%2C2.3606689742726985&withFares=true&numLegAlternatives=3&joinInterlinedLegs=false&preTransitModes=WALK%2CRENTAL&postTransitModes=WALK%2CRENTAL&directModes=WALK%2CRENTAL&preTransitRentalFormFactors=SCOOTER_STANDING%2CBICYCLE&postTransitRentalFormFactors=SCOOTER_STANDING%2CBICYCLE&directRentalFormFactors=SCOOTER_STANDING%2CBICYCLE&maxMatchingDistance=250&maxPreTransitTime=1800&maxPostTransitTime=1800&maxDirectTime=3600).

### DIFF
--- a/include/nigiri/routing/search.h
+++ b/include/nigiri/routing/search.h
@@ -195,8 +195,6 @@ struct search {
                    q_.transfer_time_settings_,
                    algo_state)},
         timeout_(timeout) {
-    utl::sort(q_.start_);
-    utl::sort(q_.destination_);
     q_.sanitize(tt);
   }
 
--- a/src/routing/direct.cc
+++ b/src/routing/direct.cc
@@ -47,12 +47,12 @@ std::optional<journey::leg> lookup_offset(location_idx_t const loc,
     return std::optional{make_leg(td->first, td->second.transport_mode_id_)};
   }
 
-  // Search for shortest offset.
+  // Search for shortest offset, assuming offsets are sorted ASC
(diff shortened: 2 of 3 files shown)


include/nigiri/routing/search.h
```diff
@@ -195,8 +195,6 @@ struct search {
                    q_.transfer_time_settings_,
                    algo_state)},
         timeout_(timeout) {
-    utl::sort(q_.start_);
-    utl::sort(q_.destination_);
```
felixguendling, line 199: they were already sorted here, guaranteeing library users like MOTIS can't make mistakes? why remove it here and put this task on the library user?

felixguendling: ah this is search.h - we can move this to query.cc to make it also work for pong

PR #398, opened 2026-08-13
Pong minlookahead

mority:
adds a minimum lookahead to Pong to prevent a query-side short max. travel time from causing Pong to miss journeys


src/routing/raptor/pong.cc
```diff
@@ -261,7 +261,7 @@ routing_result pong(timetable const& tt,
       ping.add_start(s.stop_, s.time_at_stop_);
     }
     auto const worst_time_at_dest =
-        start_time + (kFwd ? 1 : -1) * kMaxTravelTime;
+        start_time + (kFwd ? 1 : -1) * (q.max_travel_time_ + kMinLookAhead);
```
felixguendling: this effectively makes the bound 5+1=6 so we should have this change benchmarked?

mority: Exactly, that is the default case in which `q.max_travel_time_ == kMaxTravelTime == 5_days`.

We could get around benchmarking by doing 
`std::min(q.max_travel_time_ + kMinLookAhead, kMaxTravelTime)`
Would resolve the issue were PONG currently fails due to a low `q.max_travel_time_` but not risk performance degradation in the default case.

felixguendling: The current version would allow us to get writ of this:
https://github.com/motis-project/nigiri/blob/master/exe/benchmark.cc#L103

But it's totally fine to have that I think. Either way, comparison PONG vs Range will never be perfect: either we have a hack in pong.cc or in benchmark.cc

mority: Rather have the hack in benchmark.cc then. I changed it to
`std::min(q.max_travel_time_ + kMinLookAhead, kMaxTravelTime)`

PR #402, opened 2026-08-27
amd gpu support

mority:
not sure if we want this, it's a lot of glue/compat layer

but i am interested if it still passes CI on an nvidia gpu


include/nigiri/gpu_compat.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+// Replaces cista/cuda_check.h's CISTA_CUDA_COMPAT, which keys off
```
felixguendling: this should go into cista then

src/routing/gpu/raptor.cu
```diff
@@ -1176,18 +1175,18 @@ void gpu_raptor<SearchDir, WithBounds>::next_start_time() {
...
-  thrust::fill(thrust::cuda::par.on(state_.impl_->stream_),
-               begin(state_.impl_->prev_station_mark_),
-               end(state_.impl_->prev_station_mark_), 0U);
-  thrust::fill(thrust::cuda::par.on(state_.impl_->stream_),
-               begin(state_.impl_->station_mark_),
-               end(state_.impl_->station_mark_), 0U);
-  thrust::fill(thrust::cuda::par.on(state_.impl_->stream_),
-               begin(state_.impl_->route_mark_), end(state_.impl_->route_mark_),
+  thrust::fill(par_on(state_.impl_->stream_),
+               state_.impl_->prev_station_mark_.begin(),
```
felixguendling: not sure if `begin` / `end` change is necessary?

CMakeLists.txt
```diff
@@ -1,7 +1,34 @@
 cmake_minimum_required(VERSION 3.22)
 project(nigiri)
 
-option(NIGIRI_CUDA "enable CUDA functions" OFF)
+option(NIGIRI_CUDA "enable GPU routing on NVIDIA (CUDA)" OFF)
+option(NIGIRI_HIP "enable GPU routing on AMD (HIP/ROCm)" OFF)
+set(NIGIRI_HIP_ARCH "" CACHE STRING
```
felixguendling: for CUDA we explicitly set no default so the user has to set the arch string - I would prefer this behavior for AMD as well (no magic, easier for cross-compiling)

include/nigiri/routing/gpu/gpu_runtime.cuh
```diff
@@ -0,0 +1,106 @@
+#pragma once
```
felixguendling, line 1: ROCm has no compatibility layer for this?

include/nigiri/routing/gpu/gpu_std.cuh
```diff
@@ -0,0 +1,94 @@
...
+
+namespace nigiri::routing::gpu {
+
+// std::span cannot be used in HIP device code: libstdc++'s accessors reference
+// __glibcxx_assert_fail on their constant-evaluation path, and clang rejects
+// the reference to a __host__ function from __host__ __device__ code even
+// though the branch is dead at run time. This is the subset the GPU sources
+// need, annotated for the device.
+template <typename T>
+struct d_span {
```
felixguendling, line 25: ROCm doesn't provide this in the cuda namespace?

include/nigiri/gpu_compat.h
mority: cf. felixguendling/cista#282

src/routing/gpu/raptor.cu
mority: rocThrust hat kein unqualified begin/end, hab es mit std::begin/std::end ersetzt, dass es wieder vorne sitzt

CMakeLists.txt
mority: done

include/nigiri/routing/gpu/gpu_runtime.cuh
mority: this is the compatibility layer. we could run HIP on NVIDIA but not the otherway round. This file lets us keep the CUDA code. we could rewrite everything to HIP which would work on AMD and NVIDIA but we also would need HIP installed on both AMD and NVIDIA machines, in that case.

include/nigiri/routing/gpu/gpu_std.cuh
mority: ROCm 6.x would but RX580 can not go newer than ROCm 5.7

CMakeLists.txt
```diff
@@ -1,7 +1,34 @@
...
+        "AMD GPU architecture(s) to build for, e.g. gfx803 (default: autodetect)")
+
+if (NIGIRI_HIP AND NOT DEFINED CMAKE_HIP_ARCHITECTURES)
+  if (NIGIRI_HIP_ARCH)
+    set(CMAKE_HIP_ARCHITECTURES ${NIGIRI_HIP_ARCH})
+  else()
```
felixguendling: "else" won't happen and we can just use `CMAKE_HIP_ARCHITECTURES` instead of inventing our own

PR #417, opened 2026-09-18
actually apply seq encoding

traines-source:
or am I missing sth? on the other hand if leaving it out doesn't create noticeable ram usage maybe we shouldn't do it?


felixguendling (approved):
did you measure memory usage? the actual encoding should probably reduce memory usage a bit?

traines-source:
No. Can do that. But I would be more worried that this breaks something somewhere, I'm not sure how good our test coverage is for this.

felixguendling:
as long as we use the stop seq decoding everywhere, that should be fine

maybe Claude can check this?

traines-source:
Checks returned nothing suspicious, added a test over in https://github.com/motis-project/motis/pull/1605
tt size on Germany is 526 MiB vs 636 MiB, 17.25% of the total timetable file, i.e. it is worth it

